PR #4, opened 2018-05-02
correct issues in files so build test will pass without issue

dman776:
corrected a few issues in these files so that the build tests will pass without warnings

### DIFF
--- a/files/usr/local/bin/aimer
+++ b/files/usr/local/bin/aimer
@@ -22,8 +22,8 @@ else
 }
 
 
-$lastset=0;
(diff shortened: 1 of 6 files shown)


ae6xe (commented):
For the (un-referenced) variables on lines commented out,  should we delete them?   Not seeing any value to preserve the history since any code that used this variable was removed (or never existed).

dman776:
good point.  commented lines removed

ae6xe (approved):
looks good to me.

PR #5, opened 2018-05-03
fix prebuild shellcheck tests to run properly with external sources

dman776:
eliminates failure of shellcheck.  Was getting errors like:
In /opt/aredn/aredn_ar71xx/files/usr/local/bin/get_boardid line 6:
  . /lib/functions.sh
  ^-- SC1091: Not following: /lib/functions.sh was not specified as input (see shellcheck -x).


ae6xe (approved):
yes, good add.

dman776:
running ShellCheck 0.4.6 on Ubuntu 18.04 LTS (Docker container)

PR #7, opened 2018-05-03
aredn: define vlans for aredn devices

ae6xe:
aredn devices configured with lan, wan, and dtdlink
networks.


dman776:
validated with prior testing.  looks good.

PR #8, opened 2018-05-08
feature: add bootloader version to supporttool output

dman776:


ae6xe (commented):
On Mikrotik, there are 2 bootloaders (one that can be updated and one that is read only to guarantee a fall back).   Thus in the future, we'd want to test for the platform to report this additional information.   Should we template out this platform test now and only run on on ubnt out of the gate?

dman776:
not a bad idea.  Is there an existing mechanism to show the manufacturer?   /usr/local/bin/get_boardid shows it as part of the string, but, i don't trust that across all the manufacturers.  Looking for something that returns: ubnt, tplink, mikrotik, etc.

dman776:
examples of get_boardid output:
TP-Link CPE210 v1.0
Ubiquiti Nanostation M

if I had to, I could parse the up to the first space to get the mfg.

ae6xe:
check on a device in the /etc/board.json data.  If there is a unique or best object to key off, this file would tell us.  All of the downstream tools/commands, get_boardid, etc. pull from this master data.

dman776:
ok.  i added a get_hardware_mfg script that pulls/parses from board.json.  Added cmds_ubnt sections to supporttool.

ae6xe:
dman776, it installed (from a full image build process) but did not end up with 755 permissions on /usr/local/bin/get_hardware_mfg file.  Consequently, it says permission denied when I try to run it manually and there is no output in the support download (only the header).

dman776:
Hmm.  I explicitly set permissions before I committed.  At least I thought I did. Will check tomorrow 

Thanks,
Darryl

ae6xe:
then it should be good.   The issue I'm seeing must be due to how I pulled down the change in patch format, which isn't going to set the file permissions.

PR #9, opened 2018-05-09
ar71xx: Add Support for the TP-LINK CPE210 V2.0 Device

apcameron:
ar71xx: Add Support for the TP-LINK CPE210 V2.0 Device

Once Support for the TP-LINK CPE210 V2.0 gets added to the OpenWrt code we will need to drop the patch 001-add_support_for_TP-Link_CPE210_v2.patch


configs/ar71xx-generic.config
```diff
@@ -1,6 +1,7 @@
 CONFIG_TARGET_ar71xx=y
 CONFIG_TARGET_ar71xx_generic=y
 CONFIG_TARGET_MULTI_PROFILE=y
+CONFIG_TARGET_DEVICE_ar71xx_generic_DEVICE_cpe210-v2=y
```
dman776, line 4: will this be the similar for the CPE510 v2?

dman776 (commented):
do you suspect that these changes will be similar for CPE510 V2 devices?

apcameron:
I would think so but RobiMarko Would be in a better position to answer that
https://github.com/robimarko

ae6xe:
I have tested this PR = success.  live cpe210v2 running just fine browsing the mesh.   One minor change is needed for the build itself.  There is a postbuild test script that counts the number of images that are supposed to be produced.  It now fails with an error code at the very end of the build.   This is in the file scripts/tests/postbuild/700* .   There is a variable that needs updated to account for the new images created.  "NUMBEROFIMAGESCOUNT=11" needs to be 12 now.  I'm thinking we should do a separate commit for this, so when we eventually revert this commit (when it is part of openwrt), all is well.

ae6xe:
One more thought, there are more changes in this commit that we wouldn't be able to simply revert this commit as is in the future.   But it is straight forward to add a new commit that removes a couple patch files and updates the series file.   So, trying to separate openwrt unique and aredn unique changes into different commits has limited value. If we just add the test script update to this PR, I'm good to pull though.

dman776:
I prefer to bundle the numberofimagescount change in with this PR.

apcameron:
I have added a second commit to update the counter

ae6xe:
One additional issue I just discovered.  The wifi and dtdlink interfaces are being assigned the same IP address.  Everything is working as I am accessing the dtdlink interface, but I have no links on wifi.

apcameron:
What needs to be done to fix that as I do not update any code on that side as far as I know? It may be a problem for devices that do not have a switch.
Is it something that needs updating in 708-define-aredn-vlans.patch
        wp543)
-               ucidef_set_interface_lan "eth0"
+               ucidef_set_interfaces_lan_wan "eth0" "eth0.1" 
+               ucidef_set_interface_raw "dtdlink" "eth0.2" "static"
                ;;

ae6xe:
look in perlfunc.pm for the variable $dtdmac .    The IP is derived from the MAC address of an interface.   the primary hostname/IP  is derived from wifi or wlan0's mac address.  The dtdlink's IP address is derived from the lan or eth0's mac address.     The MAC of both these interfaces is the same on my cpe210v2 device.   I have a vague memory that this may be the same situation with the prior tplink devices and their might be code somewhere that sets that MAC address of the LAN to be different.   If you look on a ubiquiti device one of the MAC octets is incremented by 1 from wifi to lan MAC address.  The same pattern can be applied to tplink and manually set the MAC address.   Alternatively we update perlfunc.pm checking for tplink to do this +1.

apcameron:
/etc/init.d/local is supposed to take care of this. Does it work for other TP-LINK devices that do not have a switch?

ae6xe:
All is well on my cpe510v1.1 device with a switch.  This cpe210v2.0 doesn't have a switch.  It looks like the /etc/init.d/local code is working as the MAC address on (wifi or wlan0) and (lan or eth0) are different and what they should be.   Ether the timing of this procd local service is not happening soon enough, or the logic to set the IP addresses is not working correctly for these MACs on this device, given the IP addresses are the same on wifi and dtdlink.  Are you getting the same behavior on your cpe210v2?    "ifconfig wlan0" and "ifconfig eth0.2" both have the same IP address?

apcameron:
I have updated /etc/init.d/local to work with the TP-LINK CPE210V2.0 based on ae6xe's suggestion and tested it on my CPE210V2.0 and CPE210V1.1

apcameron:
I have squashed the 3 commits into one commit with an corrected comment

apcameron:
The series file has been updated after I synced with the upstream branch

apcameron:
Previous commits dropped and a single commit added.

ae6xe (approved):
Looks good.

PR #10, opened 2018-05-13
hardware: add tplink cpe510 v1.1 aredn device definition

ae6xe:
Power settings at or below vendor max specs.  Some channels
further limited within driver likely from vender
flash config data settings


ae6xe:
forgot to branch when making this change in my local repo.  I can do the right process next round if we can not squash these when accepting for this PR.

dman776:
this looks acceptable to me.

PR #11, opened 2018-05-14
feature: convert status page to lua to make initial page load less intensive

dman776:
convert status page to lua.
removed CSS styling for now as I cannot find an efficient mechanism to do multipart post parsing at this time (without LUCI).


dman776:
oops.  i need to enable libubus-lua module.  will update shortly.

dman776:
ok.  not sure how that got missed, but, i added that function to the utils.lua file.

PR #12, opened 2018-05-16
ar71xx: Add Support for the TP-LINK CPE210 V2.0 Device

apcameron:
Once Support for the TP-LINK CPE210 V2.0 gets added to the OpenWrt code we will need to drop the patch 001-add_support_for_TP-Link_CPE210_v2.patch


apcameron:
New Pull request created to Add Support for the TP-LINK CPE210 V2.0 Device including the missing uci commit from the earlier PR

ae6xe:
tested with the following:
1) interfaces all receive correct IP on firstboot and after basic setup
2) browse network and access remote notes and internet gateway
3) image loads with tftp and via tp-link Pharos OS upload

PR #13, opened 2018-05-19
bugfix: sysupgrade no longer supports -d option

ae6xe:
Tested on UBNT NS M5 XW and CPE210v2.0.


dman776 (approved):
makes logical sense to me.

PR #17, opened 2018-06-19
ar71xx: Add Support for the TP-LINK CPE510 V2.0 Device

apcameron:
Once the PR for this is accepted by Openwrt we will need to remove the file
001-add_support_for_TP-Link_CPE510_v2.patch

Update Code for the CPE210 V2.0 to the latest PR #937 from Openwrt

Update 99_setup_aredn_include to remove unused Rssi Led configuration


apcameron:
Removed 001-add_support_for_TP-Link_CPE210_v2.patch as PR #937 has been committed

apcameron:
I have tested this on both by TP-LINK CPE510 V2.0 and my TP-LINK CPE210 V2.0 after removing the 001-add_support_for_TP-Link_CPE210_v2.patch and it compiles and runs correctly on both devices

dman776 (commented):
Seems reasonable to me.

ae6xe:
A patch fails to apply on the 'v18.06.0-rc1' openwrt tag and upcoming release.   It does apply and builds on the openwrt-18.06 current branch.   If we wanted to do an RC on top of the 18.06.0  release, then some changes are necessary.  Let's discuss our timeline.

dman776:
What's different between 18.06 and 18.06 RC1?

ae6xe:
"18.06.0-rc1" = an existing tag, fixed point in time, of RC1 and expected
to become the final release 18.06.0.
"openwrt-18.06" = an existing branch, moving target, for creating RC,
release, .1, .2, ... point releases
"18.06.0"  = a tag, final release for 2018, has been postponed a week from
June 28.
"18.06.1"  = planned point release one to two months after release

The current builds are based on the moving target branch and we need to
determine which tag or release to lock to.  This PR fails to apply the
patches on 18.06.0-rc1.

To paraphrase in the words of the guys doing this release, the timing is
basically arbitrary given the massive contributions from the linux kernel
and 100s to 1000+ of packages by so many different groups and different
timelines that go into the release.  It's more or less, setting an
arbitrary time, so all groups know to complete work in the timeframe to get
into the release.   They will have one release a year, with 2 to 4 point
releases one to two months in succession as needed.

apcameron:
One thing to note is our current build is based on 18.06 per the openwrt.mk file
OPENWRT_COMMIT=openwrt-18.06 

It is not based on 18.06.0-rc1 yet.

ae6xe:
openwrt-18.06 is past the 18.06.0-rc1 point in time, working towards the .1
point release (or possibly if rc1 has egregious errors, will become the
release instead of the rc1).

apcameron:
@ae6xe What do I need to change to compile against 18.06.0-rc1?
Also after you have had a chance to discuss with the rest of the team let me know which build to target and I will revise the code accordingly

apcameron:
Leave a comment

apcameron:
I am creating a new request

PR #18, opened 2018-06-22
build: README add device status and openwrt URL correction

ae6xe:
See  http://github.com/ae6xe/aredn_ar71xx for the formatted output of READM.md to review.


dman776:
nice work

dman776 (approved):
looks good.  We can add the docker instructions in a new commit.

PR #23, opened 2018-06-30
stacked changes A) aredn: iperf fw rules and fw restart on pkg install; B) add definition for Basebox 2

ae6xe:
Port iperf fw rules and fw restart on pkg install to
new build structure


dman776:
These should be 2 separate PR’s

ae6xe:
Yea, thought I'd get a 2nd PR, but when you stack, stays with first.  how to undo?    rebase and merge will keep the separate commits.  I will avoid staking in the future.

PR #27, opened 2018-07-04
ar71xx: Add Support for the TP-LINK CPE510 V2.0 Device

apcameron:
Once the PR for this is accepted by Openwrt we will need to remove the file
001-add_support_for_TP-Link_CPE510_v2.patch

Removed 001-add_support_for_TP-Link_CPE210_v2.patch as PR #937 has been committed

Updated 99_setup_aredn_include to remove unused Rssi Led configuration


ae6xe:
reviewed code, built, and tested on cpe510v1.  All looked good.    k5dlq has also tested the same.

PR #33, opened 2018-07-06
dnsmasq: upgrade to dnsmasq 2.80

ae6xe:
bring dnsmasq current to what has been released in openwrt 18.06.0
dependency on nettle upgrade also to what is in openwrt 18.06.0


ae6xe:
These patches transform the package/network/services/dnsmasq definitions to exactly what is in the latest openwrt 18.06.0 release.  the crypto package nettle had dependencies and required an upgrade.  Also, the identical dnsmasq aredn init script is reused (identical init as in 3.16.1.1 and current develop branch to be compatible with olsrd).

dman776 (approved):
looks ok.

PR #41, opened 2018-07-09
bugfix: node cannot detect that it was WAN access

### LINKED ISSUE #40: node not online (/usr/sbin/ip location is incorrect)

dman776:
/usr/sbin/ip command is incorrect (fixes #40)

### DIFF
--- a/files/www/cgi-bin/perlfunc.pm
+++ b/files/www/cgi-bin/perlfunc.pm
@@ -462,20 +462,20 @@ sub get_default_gw
     my $gw = "none";
 
     # Table 31 is populated by OLSR
-    foreach(`/usr/sbin/ip route list table 31`)
-    {                                          
+    foreach(`ip route list table 31`)
+    {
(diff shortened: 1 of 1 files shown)


ae6xe (commented):
strictly speaking, there's one more condition for this table to be in the routing path.  it must be in the preference list as reported by "ip rule".  There's a new feature in the pipeline to block internet traffic routing though a node, by removing this table from the chain of tables to determine a route.  The table 31 will may still exist, but not be referenced/used.   We need to add this 2nd condition here, ether in this commit or the next commit for this new feature.     These are the rules that would be removed to test for as the 2nd condition:
ip rule del pref 20080
ip rule del perf 20080
ip rule del pref 30099
ip rule del pref 30299

dman776:
i think since those are related to the new feature, then, they should be grouped into the new commit.

ae6xe:
Ok, go ahead and commit.

PR #43, opened 2018-07-10
bugfix: correct config.mk package path

dman776:


ae6xe (approved):
Looks correct.  The target/subtarget path is now encoded in the directory structure below this level under openwrt/bin .

PR #50, opened 2018-07-11
circleci: add workflow to build/deploy nightly snapshots at 2100UTC

dman776:
enabled workflows and distinct jobs.  
Nightly build will run at 2100UTC
Added deployment of readme.md
Added generation of changelog.md

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,9 @@
 version: 2
 jobs:
+  ######################################
+  # Common
+  ######################################
   build:
-    branches:
-      only:
-        - develop
     docker:
       - image: arednmesh/builder
     steps:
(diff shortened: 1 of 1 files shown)


dman776:
The Makefile change to remove the deletion of the openwrt/bin directory should precede this commit.

ae6xe (approved):
I'm not setup to test, but logic looks good.

ae6xe:
This is 2PM CST.   If we typically make changes in the evening, we wouldn't see a build until late afternoon.   If this was at 2AM CST, after we've gone to bed, then the build would be ready in the AM.

dman776:
changed to 0900 UTC

ae6xe (approved):
looks good.

PR #58, opened 2018-07-12
bugfix: signal charts drop-down list (fixes #48)

### LINKED ISSUE #48: Charts function not fully functional
In nightly release aredn-develop-21-develop-360786c  the only option available under Charts is
"Average signal for all connected stations"

dman776:
this is a partial fix.  The drop-down will now populate with IP address.
There remains an issue with name resolution so that the drop-down will contain names (when available).


ae6xe (approved):
logic looks good.

PR #60, opened 2018-07-13
circleci: add proper dependencies to control order of workflow exec

dman776:


ae6xe (approved):
not too familiar with circleCI .yml yet, but looks good to me.

PR #63, opened 2018-07-14
circleci: correct proper targets directory structure

dman776:


ae6xe (approved):
That looks correct, both packages and targets dirs are put at the trunk level.

PR #73, opened 2018-07-16
Support more direct hosts

### LINKED ISSUE #49: Not enough DHCP addresses available in some circumstances
In some cases, the 13 available DHCP IP addresses from a node are not quite enough. This feature request is for supplying up to 29 IP addresses from a node.

jbrady42:
Closes #49 

There is a discussion to be had around the chance of address collision and how best to deal with that long term. For the time being, this should give users plenty of breathing room with addresses in direct mode.

### DIFF
--- a/files/www/cgi-bin/setup
+++ b/files/www/cgi-bin/setup
@@ -131,7 +131,7 @@ $lan_proto = "static";
 # enforce direct mode settings
 # (formerly known as dmz mode)
 $dmz_mode = 2 if $dmz_mode != 0 and $dmz_mode < 2;
-$dmz_mode = 4 if $dmz_mode > 4;
+$dmz_mode = 5 if $dmz_mode > 5;
 
 if($dmz_mode)
 {
@@ -398,7 +398,7 @@ if($parms{button_save})
     {
 	push @errors, "password must be changed during initial configuration";
     }
-	
+
     if($nodetac =~ /\//)
     {
 	$nodetac =~ /^\s*([\w\-]+)\s*\/\s*([\w\-]+)\s*$/;
@@ -729,6 +729,7 @@ selopt("NAT", 0, $dmz_mode);
 selopt("1 host Direct", 2, $dmz_mode);
 selopt("5 host Direct", 3, $dmz_mode);
 selopt("13 host Direct", 4, $dmz_mode);
+selopt("29 host Direct", 5, $dmz_mode);
 print "</select></td>\n</tr>\n";
 push @hidden, "<input type=hidden name=lan_proto value='static'>";
 


ae6xe:
jbrady42, these are the right changes.  Thanks for taking initiative.  There's more to the story:

1) code in /www/cgi-bin/setup that limits dmz_mode to 4 -- a check to increase before this will work.
2) just after this check is the algorithm to derive the IP addresses from a MAC address.   We need to do the engineering-math to determine if this additional allocation will be as close to random as we can get.   Someone buys 2 devices at the same time with a mac address +1 apart.   Fundamentally, will we have confidence the algorithm as-is will interleave the subnet allocations sufficiently, such that the chance of collision is not a concern (or if it is a concern, how big of concern).  For sure, we need to make sure the mac+1 common case isn't an instant collision.

ae6xe:
I reviewed the algorithm.  There's no issue for successive MAC addresses.  What happens is the derived 10.x.x.x (decimal number of the) wifi address is multiplied by 4, 8, 16, and now 32. to spread out the starting addresses of the LAN subnet.  Then the value is AND'd with 0xFFFFFF to get the new x.x.x part.  This gives it some randomization.  
 
If everyone used a 29 address LAN, then the max nodes that would fit is just over 524,000 nodes.  Although, we'd expect to start seeing collisions probably in the 10,000+ range.  There are other scale issues however, e.g. the volume of traffic for a mesh's routing protocol of just a few 1000 nodes would significantly clog the network before we'd see an IP collision due to the number of nodes., particularly with OLSR. 

While the probability of collision will go up, my view is this is an acceptable % because we'd run into network scaling issues before there's a meaningful count of nodes this would become an issue.    Also, if there is a collision (and we've not seen one to date), one randomly changes the wifi IP address to something else and the LAN subnet will change accordingly.  It's a relatively simple fix.  

Never-mind on the 5 limit, you already have it.  I think we can commit this as is.   Are you running this live already?

jbrady42:
I have flashed this to a NSM2 and everything looks good from the admin side, and on a small test net, but don't have anywhere near enough lan devices or ports to test it under normal circumstances. Regarding the address space, yes that would leave 2^19 or the 524k you referenced if everyone used this size. Hopefully we can rely on not everyone exposing nets this large. Also agree that there would probably be some bigger issues like the routing daemon once the network approaches the size where collisions come in to play. I can do some more testing with grabbing the full allocation on virtual devices if needed, otherwise this is probably good to go.

ae6xe:
I tested this on a node and all worked as expected.  the interface was correctly configured and the port forward page worked correctly.  CDIR notation shows correct on status page.     This is not new, but when changing the LAN count,  I was reminded that the IP reservations and service advertisements have to be redone.   Let's let this sit overnight to give others a chance to comment before committing.

dman776:
When I set 29 hosts, it sets DHCP start at 2 and DHCP end at 30.  That's more than 29 addresses.

dman776:
tested ok for me.

dman776 (approved):
I'm good with this change.

dman776:
@jbrady42, what is your email and callsign (so I can attribute your work in our CONTRIBUTORS file)?

PR #77, opened 2018-07-17
feature: nodename field on setup page is too small

### LINKED ISSUE #76: Node Name textbox on Basic Setup screen is too small
Increase the size of the Node Name textbox on the Basic Setup screen so you can see the entire name of nodes with longer names. 

This shows the approximate size right now compared to the typical length of node names:
aredn_textbox_size_mesh_status

dman776:
fixes #76


ae6xe (approved):
looks good.

PR #78, opened 2018-07-17
Added the Nodes locally hosted service listing to sysinfo.json

r1de:
Added the Nodes locally hosted service listing to sysinfo.json file.

Yes, another addition for my mapping system.
The only way to get this info before was port 9090 and we know how that went. :)

### DIFF
--- a/files/www/cgi-bin/sysinfo.json
+++ b/files/www/cgi-bin/sysinfo.json
@@ -268,7 +268,7 @@ if string.find(nixio.getenv("QUERY_STRING"):lower(),"services_local=1") then
 		service['error']="Cannot read local services file"
 		table.insert(lclsrvs, service)
 	end
-	info['local_services']=lclsrvs
+	info['services_local']=lclsrvs
 end
 
 -- Output the HTTP header for JSON
--- a/files/www/cgi-bin/sysinfo.json
+++ b/files/www/cgi-bin/sysinfo.json
@@ -55,7 +55,7 @@ end
 info={}
 
 -- API version
-info['api_version']="1.2"
+info['api_version']="1.3"
 
 -- NODE name
 css=getUciConfType("system", "system")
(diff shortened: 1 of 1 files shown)


files/www/cgi-bin/sysinfo.json
```diff
@@ -244,6 +244,33 @@ if string.find(nixio.getenv("QUERY_STRING"):lower(),"services=1") then
 	info["services"]=services
 end
 
+-- LOCALLY HOSTED SERVICES ONLY
+if string.find(nixio.getenv("QUERY_STRING"):lower(),"services_local=1") then
```
dman776, line 248: I would prefer "local_services" instead in order to match the info[] table name

r1de: I get what you are saying now. Sorry. :smile:
I made this "services_local=1" due to the fact that "local_services=1" will *also* trigger the "services=1" function.
I wasn't looking for *all* of the services, just the locally hosted ones, and felt that triggering the other function at the same time was a big waste of bandwidth.
I am open to other names for it. I just used those so they would make sense.

r1de: I could just swap the info[] line to read services_local instead. :smile:

files/www/cgi-bin/sysinfo.json
```diff
@@ -244,6 +244,33 @@ if string.find(nixio.getenv("QUERY_STRING"):lower(),"services=1") then
...
+				service['protocol']=protocol
+				service['link']=link
+				table.insert(lclsrvs, service)
+			end
+		end
+	else
+		service['error']="Cannot read local services file"
+		table.insert(lclsrvs, service)
+	end
+	info['local_services']=lclsrvs
```
dman776: agree, for now, just rename the info table to info['services_local']

r1de:
Done

PR #86, opened 2018-07-20
Format Code

jbrady42:
I started going through to code to get acquainted in preparation for some changes, and noticed there were a lot of places with inconsistent format or code had been dropped in with different indentation. Ran some of the perl files through the basic formatter in atom editor to clean everything up.

This promotes readability and long term maintainability as it helps people get up to speed faster on the code and is just more pleasant to work with. 

Had to do some minor manual tweaks in a few places, but its exactly the same code. I am testing this on a node and everything looks good so far.

### DIFF
--- a/files/usr/local/bin/aimer
+++ b/files/usr/local/bin/aimer
@@ -6,19 +6,19 @@
 
 
 sub aimmode_printusage{
-   print "USAGE:\n";
-   print "$0 SSID\n";
-   die;
+  print "USAGE:\n";
+  print "$0 SSID\n";
+  die;
 }
 
 
 # SSID NAME
 if ( $ARGV[0] ) {
-    $SSIDName= lc $ARGV[0];
+  $SSIDName= lc $ARGV[0];
 }
 else
 {
-    $SSIDName= lc "AREDN-20-v3";
+  $SSIDName= lc "AREDN-20-v3";
 }
 
 #dbm values for each RSSI trigger level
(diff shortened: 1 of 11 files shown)


dman776:
This reminds me.  We need to remove the aimer script.  It was experimental and no longer needed since has real-time charts

ae6xe:
do you want to do more burn in testing, or shall we move this into a nightly build?  Looks good to me.

dman776:
I’m good with it

Thanks,
Darryl

jbrady42:
Just tried on a different node and seems to be good. I have been having some intermittent issues with the OTA upgrade. Have had to tftp a few times after an attempted upgrade, but it happens with the official nightlies as well so I'm thinking it's an iffy device. Certainly will not hurt to give it a quick test if yall want to be sure. Another thing I noticed is the MCS and T\RX Rate on the chart are missing, but looks like that is an existing issue in the nightly as well.

ae6xe:
What devices are you seeing intermittent issues with?  (32Mb RAM devices are unstable and don't have resources to do sysupgrade.  Need to do an effort to remove RAM consumption that has plagued all nightly builds since 3.16.1.1).

jbrady42:
My main testing device is a NS M2 XW, so it should have enough memory.

Excerpt from debug info:
```
========== cat /proc/cpuinfo ==========
system type		: Atheros AR9342 rev 2
machine			: Ubiquiti Nanostation M XW
processor		: 0
cpu model		: MIPS 74Kc V4.12
BogoMIPS		: 266.64
wait instruction	: yes
microsecond timers	: yes
tlb_entries		: 32
extra interrupt vector	: yes
hardware watchpoint	: yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa			: mips1 mips2 mips32r1 mips32r2
ASEs implemented	: mips16 dsp dsp2
shadow register sets	: 1
kscratch registers	: 0
package			: 0
core			: 0
VCED exceptions		: not available
VCEI exceptions		: not available

========== cat /proc/meminfo ==========
MemTotal:          60188 kB
MemFree:           31288 kB
MemAvailable:      26512 kB
Buffers:            3384 kB
Cached:            10480 kB
SwapCached:            0 kB
Active:            12456 kB
Inactive:           4712 kB
```

jbrady42:
Just tested on a PBE-M2-400 with zero issues, so I'm fairly sure it's a problem with my other node.

ae6xe (approved):
looks good.

PR #88, opened 2018-07-21
ar71xx: Add Support for the TP-LINK CPE210 V3.0 Device

apcameron:
This Device is almost the same as the CPE210 V2.0
Once Support for the TP-LINK CPE210 V3.0 gets added to the OpenWrt code we
will need to drop the patch 001-add_support_for_TP-Link_CPE210_v3.patch


ae6xe (changes requested):
Andrew,   I don't think this value is functionally used, but to avoid future confusion, can you set the "maxpower = 29" to be "25"?

apcameron:
Maxpower has been updated to 25 as requested

ae6xe (approved):
looks good.  I don't have a v3 to test with.

Issue #85, opened 2018-07-20
nslookup error


Orv:
On Nightly Build 59, nslookup works, but throws this error, which may be causing other issues we've seen (not ready to talk about those yet):

root@WD6EBY-LA-MtWilson-SW-Sector-5G:~# nslookup w6bi-vc-qth-2g
Server:         127.0.0.1
Address:        127.0.0.1#53

Name:      w6bi-vc-qth-2g
Address 1: 10.62.29.243
***** Can't find w6bi-vc-qth-2g: No answer**
root@WD6EBY-LA-MtWilson-SW-Sector-5G:~#

ae6xe:
The code for the operation of the node is using nslookup to do the reverse ip-to-hostname lookup.  (At least that's all I found when searching all the code.)   While you might get this from the command line, hostname-to-ip, the output wouldn't affect the node operations since it's not used.  When ready, post any failures observed.

Orv:
Keep this open for now; Doug, WB2YXY, who uses bind and dnsmasq together, reports that the bind response back from the nightly builds differs from the (correct) response back from the 3.16 production release.  Hope to have more coherent information soon.

Orv:
We can close this; we won't be pursuing it for now.

PR #96, opened 2018-07-28
Trim Packages

r1de:
Kernel support for USB has been moved to modules
libpcap and tcpdump have been moved to modules
libnetsnmp and snmpd have been moved to modules
IPv6 has been removed from the kernel
IPv6 support has been removed from busybox network utils as well.
(no more funny DNS responses in nslookup! Remember issue #85 ?)

Please take a look, some of the options for IPv6 and busybox are probably redundant.
Image file is at ~~5.7MB~~ 5.6MB with this config.


r1de:
There might be some other things that can be removed from busybox as well.
"dd", do we really need "dd" on the node?
"ls" with colors, sure it's nice, but is it really needed?
"chroot", etc...etc...

r1de:
Additional patch file to stop busybox from making ping6 and traceroute6 since they'd no longer be needed.
Adding CONFIG_BUSYBOX_CONFIG_PING6=n into common.config did not seem to do anything.
Busybox had to be forced to default to not creating ping6 and traceroute6.

dman776:
If I recall, i think the dd command is used in the get_hardware (or get_model) script.  That needs to remain.

dman776 (commented):
Need to re-number the busybox patch to proper numbering structure (check the readme.md file).
also, need to add it to the patches/series file so that the build will apply it.

ae6xe (commented):
[edit]   nevermind, I see you already tried this, should have read the full history :) first.  

Please see if you can instead  edit the configs/common.config and specify what the setting should be:

CONFIG_BUSYBOX_DEFAULT_TRACEROUTE=n
CONFIG_BUSYBOX_DEFAULT=TRACEROUTE6=n

Probably a patch to openwrt is not needed, rather a config file setting.

ae6xe:
Yes, 'dd' is used in a few places to dig though flash memory looking for specific memory locations to pull out settings or identify the device.

ae6xe:
note, this removal of IPV6 was done once before, just not ported forward yet.  Darryl, I think you did that, so assume the changes here are equal to or greater than what was done before.  

I ran into a global compile option in "make menuconfig" a couple months ago to exclude compiling in code to all the packages/models/etc related to ipv6.   Is this how the options in this PR were determined?  Could there be more options to include?

r1de:
@dman776 Thanks! I will look at the numbering, I just put 999 so as to not get in the way of the others.
Once I created the patch file, it seemed like it got picked up at build time without me doing anything, but I will check that too.

I will forget about removing "dd".
You guys got me thinking of how to really slim it down and I might get over zealous a bit, sorry.
I come from an old line of people who, even still, build our own kernels for most of our linux machines, I know, I'm crazy, it's fun! :smile:

@ae6xe I determined the needed options from looking at both the .config file and from the "Help" area of "make menuconfig", and it seems they are sometimes different, and like busybox, some seem to work and others (that you'd think should) won't.
I could have been having a PEBKAC issue though. :smile: 
*edit*: and yes Joe, there are probably many other options that could be used to clean things up a bit, remove certain commands that no one will need (or get into trouble with), etc..etc...

I also think the order might matter in common.config, but I haven't really determined that to be the case yet... It's just a hunch.

dman776:
the build seemed inconsistent, but, not sure if it is related to this PR or not.  I did need to 'make' twice to get this to build completly (using the official "clean" docker container).  Loaded on Rocket M2 successfully.
Loaded SNMP packages.  Looks like the /etc/config/snmp is different now.

--------
Installing snmpd (5.7.3-9) to root...
Collected errors:
 * resolve_conffiles: Existing conffile /etc/config/snmpd is different from the 
conffile in the new package. The new conffile will be placed at
/etc/config/snmpd-opkg.
Configuring snmpd.
------

dman776 (approved):
Tested ok on Rocket M2.

ae6xe (approved):
looks good.

PR #101, opened 2018-08-02
bugfix: node description not saving properly

### LINKED ISSUE #89: Node Description improvement #1
Node Description is wiped out across firmware upgrades - can this be rectified?

dman776:
This fixes #89


ae6xe (approved):
looks OK, but unable to test for now.

PR #103, opened 2018-08-03
bugfix: iperf patch file update

r1de:
fixed version number for new package
add --disable-ipv6 to iperf configure args


dman776:
build fails due to IPv6 not being available still.

dman776:
withdrawing this PR

PR #105, opened 2018-08-03
bugfix: upgrade iperf to iperf3

dman776:
original iperf2 has issue running without IPV6
iperf3 is newer and leaner with additional features such as JSON output capability
iperf3 is slightly smaller in binary footprint


r1de:
iperf3 default port is 5201 now.
It's easy enough to change ports but no one will.

ae6xe (commented):
These patches look to  be applied at the top level of the aredn_ar71xx directory.  However, the Makefile cd's to the openwrt to apply the patches.   .../openwrt/patches -> ../../patches .   Something is odd, as this should't work as is from the openwrt dir, it shouldn't be able to find another openwrt dir.

ae6xe (approved):
looks good to me.  Traveling, unable to compile and test for now.

PR #107, opened 2018-08-04
bugfix: allow apostrophes in node description

r1de:
node description with apostrophes would not show up on status page.
now it does.
warning message about formatting added to setup page.


dman776:
suggest that the following be HTML escaped before persisting:
"
'
<
>

PR #109, opened 2018-08-04
aredn: remove duplicate file and minify

r1de:
I found a duplicated file in www, only 16k, but every bit matters now.
Removed the duplicate and replaced the file with a minified version.
Updated www/cgi-bin/scan to use new minified file.
Tested. works fine.


ae6xe (commented):
the issue is long term maintainability.   Darryl, interested in your comments -- didn't you write this?   Looks like one version is the 'maintainable' copy used to generate the mini by removing unnecessary chars for run time.   Is there a way to consolidate to the maintainable version, then at run time strip the unnecessary chars to send to the client browser?  (Is this the reason there is a 'mini')?

dman776:
We only need the minified version

Thanks,
Darryl

ae6xe (approved):

PR #112, opened 2018-08-05
ci: add checks to determine if build is required

dman776:
also save build info upon successful build
cleaned up un-needed info


ae6xe (approved):
Looks good.

PR #113, opened 2018-08-07
ci: make the make -jX parameter configurable

dman776:
Added an environment variable to CircleCI (DASH_J_VALUE)


K6AH (commented):
I can't comment on the technical worthiness of this change.  It seems innocuous enough.

PR #114, opened 2018-08-07
ci: add check2build to workflow section

dman776:


dman776:
needed to add check2build job to workflow section

PR #115, opened 2018-08-08
build: fix compilation failure for luasrcdiet

dman776:
From bb336671fff613b812b308c7eeeeda38608bda9d Mon Sep 17 00:00:00 2001
From: Jaymin Patel <Jaymin.Patel@Sophos.com>
Date: Mon, 30 Jul 2018 15:06:14 +0530
Subject: [PATCH] fix compilation failure when luasrcdiet is being copied to
 non existing directory

Signed-off-by: Jaymin Patel <jem.patel@gmail.com>

### DIFF
--- a/patches/200-luasrcdiet-luci-build.patch
+++ b/patches/200-luasrcdiet-luci-build.patch
@@ -1,0 +1,17 @@
+From bb336671fff613b812b308c7eeeeda38608bda9d Mon Sep 17 00:00:00 2001
+From: Jaymin Patel <Jaymin.Patel@Sophos.com>
+Date: Mon, 30 Jul 2018 15:06:14 +0530
+Subject: [PATCH] fix compilation failure when luasrcdiet is being copied to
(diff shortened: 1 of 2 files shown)


dman776:
wait.  the patch paths are wrong again.  not sure why my quilt is acting different.

ae6xe:
Darryl,  you had it right the first time.  patches are created in reference to the openwrt directory, the top of buildroot.

dman776:
i'm having to do a make feeds-update BEFORE a make because openwrt/feeds does not exist when the patch is attempted.

ae6xe:
Yea, the patch process doesn't currently work well with feeds, because a feed is another git repo beyond openwrt's.    it would be like trying to patch our aredn_packages repo this way, rather than making the change directly in the repo.       To resolve, I'm thinking we could change the dependency order and update feeds first, then apply all the patches.   Just wondering if any negatives of this order...

PR #117, opened 2018-08-08
build: move patch process after feeds update and clean up warning

ae6xe:


dman776 (approved):
tests ok on my local build

PR #128, opened 2018-08-09
bugfix: correct port order on airrouter

dman776:
AirRouter should have LAN on ports 1-3 and DTD on port 4.


ae6xe (commented):
I think the labels need reversed too?   The dtdlink is on external label port 4, so :4:lan:1,  3:lan: 2, ... .  we could add the ":4" should in the future we use the luci UI in some way.

ae6xe (approved):
looks good to me.  I don't have an airRouter to test.

dman776:
Tested on AirRouter and it works as expected.

PR #135, opened 2018-08-13
kernel: re-add patch for AT8032 Ethernet PHY

ae6xe:
The patch was wrongly removed by a kernel version bump to 4.9.106 in
the believe that it was merged upstream thow it wasn't. This lead to
unrecoverable link losses on devices which use those PHYs such as
many ubnt single-port CPEs.

Fixes: 6f8eb1b50f ("kernel: bump 4.9 to 4.9.106 for 18.06")
Signed-off-by: Daniel Golle <daniel@makrotopia.org>
(cherry picked from commit a497e47762)


dman776 (approved):
previously tested ok.

PR #136, opened 2018-08-13
bugfix: /tmp/hosts dir is missing

dman776:
names cannot be resolved from node since dnsmasq did not start properly.
errors accessing /tmp/hosts/dhcp since /tmp/hosts does not exist


ae6xe (approved):
looks good.  I've tested this.

PR #140, opened 2018-08-16
bugfix: fix olsr and dnsmasq upgrade inconsist config settings

ae6xe:
olsr was not triggering an update for dns to receive host
information due to different path names.  Also fixed
dnsmasq warning messages at start up with missing
/tmp/hosts directory.


dman776 (approved):
tested externally successfully

PR #142, opened 2018-08-19
config change to make uhttpd listen on port 80 and 8080

tpaskett:
its time to do this. i'm tired out having to teach new people to mesh how to type in the url and they always forget, especially during our emergency exercises. port 8080 does not make it more secure. it is also very unlikely users have the external interface using a public internet ip address. this change makes it work on both port 8080 for backwards compatibility and port 80 that most users are used to.

### DIFF
--- a/files/etc/config.ap/firewall
+++ b/files/etc/config.ap/firewall
@@ -64,6 +64,12 @@ config rule
 
 config rule
        option src              wifi
+       option dest_port        80
+       option proto    tcp
+       option target   ACCEPT
+
+config rule
+       option src              wifi
        option dest_port        2222
        option proto    tcp
        option target   ACCEPT
--- a/files/etc/config.ap/firewall
+++ b/files/etc/config.ap/firewall
@@ -63,7 +63,7 @@ config rule
        option target   ACCEPT
 
 config rule
-       option src              wifi
+       option src              wan
        option dest_port        80
        option proto    tcp
        option target   ACCEPT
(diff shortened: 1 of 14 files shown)


ae6xe (approved):
I agree.

dman776:
Do firewall rules need to change?

ae6xe:
good catch, yes need an update to the default firewall rules in the config files and in /etc/local/mesh-firewall/01-tunnels.

dman776 (changes requested):
needs firewall adjustments

ae6xe (commented):
several more places in etc/config*/firewall.

ae6xe:
also update to www/help.html of both ports.

tpaskett (commented):
someone could do better on the help text im sure.

ae6xe:
better grep 8080 to find all the occurrences.  The  etc/config*/firewall files have it in 3 places, 2 more to go  -- to be accessed coming in over wifi, wan, and dtdlink.    We got the incoming from tunnel already .

tpaskett (commented):
think i have all of them now

ae6xe:
Looks good to me.

dman776 (approved):

dman776:
looks good.

Orv:
This works fine (thanks, Trevor!).   I noticed that the broadcasted URL still advertises on port 8080. Should this be changed to 80?

dman776:
no, because other V3 protocol nodes expect them on 8080.   This is just for convenience.

PR #143, opened 2018-08-19
show the SNR above the real time chart

tpaskett:
this is very helpful when your on the tower and you have someone below reading you the SNR to optimize the tilt. hard to tell exactly from the chart other than better / worse. i coded / tested this on 3.16.1.0, so will have to test that it works with the changes that were made to www/cgi-bin/signal since 3.16.1.0

### DIFF
--- a/files/www/cgi-bin/signal
+++ b/files/www/cgi-bin/signal
@@ -153,9 +153,8 @@ $header = <<EOF;
       var updateRealtimeChart = function () {
         \$.getJSON("/cgi-bin/signal.json?realtime=1&device=$parms{device}", function (result) {
           dps[0].push(result[0][0]);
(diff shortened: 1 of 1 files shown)


dman776:
This doesn't appear to work on 3.16.2.0RC1.

dman776 (changes requested):
still fails on the lastest develop branch code.
Javascript error in signal 
dps[1] does not exist.
from line:   dps[1].push(result[1][0]);

signal.json only returns one element in that array. dps[0]

tpaskett (commented):
this works for me against nightly build 186-8d3b63a

dman776 (approved):
works fine on nightly

PR #146, opened 2018-08-22
bugfix: aredn slugbug mitigation

ae6xe:
with low memory conditions, typically on 32Mb RAM, the
device would become unresponsive in hours to days. The
symptoms only occured when no RF links. iw processes
would hang in Zombie state.  Updates to use iwinfo where
possible and avoid using both iw and iwinfo.  crontab
script is implemented to detect Zombie processes and free
up resources in the reduced chance the symptoms are still
occuring.


dman776 (approved):
Approved based on Randy/Darryl/Joe's testing on nodes previously.

PR #147, opened 2018-08-22
Add testing support for the PowerBridge M5

txcoder:


ae6xe (approved):
looks good.  inspected support data, didn't see any concerns.

dman776:
Can put in lat/lon values and submit to the AREDN map when you get a chance.  the map should show it properly as a PowerBridge M5 once you do.

PR #167, opened 2018-08-30
Fw upload area enhance

r1de:
aredn: Info about the device and firmware type needed added above the firmware uploading buttons.
Even I have to double check which firmware is for which device sometimes.
This will hopefully help to alleviate any confusion, and save a few people some sanity. :)

Also got rid of a warning in the syntax checking routines (03-cgibin_admin).

### DIFF
--- a/files/www/cgi-bin/admin
+++ b/files/www/cgi-bin/admin
@@ -88,6 +88,13 @@ $parms{dev} = 1 if -e "/tmp/developer_mode";
 # set the wget command options
 $wget = "wget -U 'node: $node'";
 
+#get device name and firmware type
+$deviceName = `/usr/local/bin/get_model`;
+$fwType = `/usr/local/bin/get_hardwaretype`;
+#strip off "station" from "nanostation"
+#to match the actual name of the fw files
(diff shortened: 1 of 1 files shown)


ae6xe:
I just checked on the Mikrotik gear.  Unfortunately,  this will show up with the device model detail:   get_hardwaretype on a basebox 5 shows 'rb-912uag-5hpnd".    on the hAP lite, the same.    The dmesg detail on mikrotik doesn't help ether.  It shows board=711Gr100.    2 options I can think of, A) don't show anything; B) create a mapping table of model # to image name to show.

r1de:
Ok. good catch Joe.
"more work to be done"
We'll show nothing for now.

I am removing the PR while I work on this some more.

Orv:
Could you not enable this PR for Ubiquiti products only, for now?  And once the Mikrotik issue is resolved enable it for them?  Does it work on TPLink?  If no, the same reasoning would follow.

ae6xe:
It's common code.  With such limited non-ubnt models, it would be minor difference between a check for these models and take corrective action vs check for these models and not do anything.

PR #174, opened 2018-09-05
bugfix: sysinfo.json not returning chanbw on hapACLite

dman776:


ae6xe (approved):
This works, looks good.

PR #175, opened 2018-09-05
bugfix: sysinfo.json returns board_id = 0x0000 for Mikrotik devices

dman776:


ae6xe (changes requested):
The following path, a few lines above, should be changed, (any other lua instances?)  to something like:
From:  /sys/devices/pci0000:00/0000:00:00.0/ieee80211
To:  /sys/class/ieee80211/

The path may also have phy0, phy1, ... in it and would need calculated, e.g. wlan0 -> phy0, wlan1 -> phy1, ...   key radio? and phy? off of master "wlan?".

"pci0000:00/0000:00:00.0" is for a wireless interface on a PCI bus.  Mikrotic has interfaces on chip in use with a different path.  There are links under /sys/class that take this path out of the name.

this change called "get_boardid", should be replaced with "get_model" since mikrotik doesn't have a boardid.

dman776:
/sys/class/ieee80211/phy0/device points to the exact same location:   /sys/devices/pci0000:00/0000:00:00.0

ae6xe:
On a Mikrotik hAP lite, the "/sys/devices/pci0000:00/0000:00:00.0" path is the 5Ghz interface on the PCI bus.   We need to get to the 2Ghz interface on phy1/wanl1/radio1 configured as the mesh wifi interface.   2 options:  A) calculate it needs to be phy1 in the path /sys/class/ieee80211/phy1 or B) calculate the device bus path /sys/devices/<wifi device path>/ .       I used option A elsewhere.   I'll send the password to directly access the hAP lite to inspect separately.  the wifi device path is saved in /etc/aredn_include/radio{0,1}*  for include in /etc/config/wireless settings.

dman776:
pulling this PR for now.

PR #183, opened 2018-09-07
bugfix: show freq on status page instead of channel number for 3GHz

ae6xe:
3GHz has no common assignment of channel numbers, given no part 15
coverage in the US.  AREDN devices show freq in place of a channel
number for 3GHz.


r1de:
Could we finally just add freq info to sysinfo.json now? (it really won't add much traffic to the mesh, I swear! :) )

ae6xe:
Works for me, but submit as a different PR,.

r1de:
Roger Roger.
It won't be until at least Thursday tho, sorry. (we're going out of town)

dman776 (approved):

PR #185, opened 2018-09-10
rework sysinfo.json to improve structure

dman776:
re-organize "info" functions into their own lib
Also changed sysinfo.json output to have a better structure.   ie.  location info is now under:
"location": {
     "lon": "-95.558470",
     "lat": "30.227114",
     "grid_square": "EM20ff"
   },
and mesh RF settings are under:
"meshrf": {
     "chanbw": "5",
     "ssid": "AREDNTEST-5-v3",
     "channel": "-1"
   },

This will require some detection of this version of the API and rework for systems using it like map generation scripts.

### DIFF
--- a/files/usr/lib/lua/aredn/http.lua
+++ b/files/usr/lib/lua/aredn/http.lua
@@ -43,6 +43,7 @@ end
 function json_header()
    print("Content-type: application/json\r")
    print("Cache-Control: no-store\r")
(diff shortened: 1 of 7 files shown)


ae6xe (approved):
1 instance and usage of "egrep" is recommended by shellcheck to change to "grep -E".

PR #188, opened 2018-09-13
aredn: harden cron maintenance scripts

ae6xe:
ensure maintenance scripts execute one at
a time and never in duplication

### DIFF
--- a/files/usr/local/bin/clean_zombie.sh
+++ b/files/usr/local/bin/clean_zombie.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-<<'LICENSE'
+true <<'LICENSE'
   Part of AREDN -- Used for creating Amateur Radio Emergency Data Networks
   Copyright (C) 2018 Joe Ayers AE6XE
   See Contributors file for additional contributors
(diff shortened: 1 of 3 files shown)


files/usr/local/bin/clean_zombie.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-<<'LICENSE'
+true <<'LICENSE'
```
dman776, line 2: what is "true" doing?

files/usr/local/bin/clean_zombie.sh
```diff
@@ -36,36 +36,43 @@ LICENSE
...
 
+zombiepid="/tmp/clean_zombie.pid"
+
+[ -e $zombiepid ] && [ -d "/proc/$(cat $zombiepid)" ] && exit
+
+echo "$$" > $zombiepid
+
 # wait for rssi_monitor and snrlog to run
-sleep 10
+sleep 20;
```
dman776, line 46: my concern is that this keeps the shell loaded in memory twice as long.

files/usr/local/bin/clean_zombie.sh
ae6xe: This resolved a shellcheck warning I ran into.  This is a shell pipe command going into nothing.

files/usr/local/bin/clean_zombie.sh
ae6xe: The shell is likely in RAM 100% (shared code) all the time.  It's function has a very lite footprint only doing a 'ps' and checking file system.   We could just remove the sleep altogether.

dman776 (approved):

PR #189, opened 2018-09-14
remove ipv6 address labels (ie. none)

dman776:
remove 'none' label from interface addresses on the status page.
these were IPv6 addresses.


ae6xe (approved):
looks good.

PR #190, opened 2018-09-14
mtd: spi-nor: Add Winbond w25q128jv support

ae6xe:


dman776 (approved):
Approving upstream patch

PR #191, opened 2018-09-14
feature: apply sideloaded firmware

dman776:
This PR provides an alternate mechanism to update firmware on marginal/slow links.
The device appropriate "sysupgrade" file should be copied (via FTP, or other mechanism that is not subject to network timeouts) to /tmp/web/sideload_firmware.bin.  
Once completed, the admin page will show an "Apply Sideloaded Firmware" button.
Click the button to process the firmware and upgrade.

### DIFF
--- a/files/www/cgi-bin/admin
+++ b/files/www/cgi-bin/admin
@@ -132,6 +132,15 @@ if($parms{button_refresh_fw})
 # and set $fw_version
 firmware_list_gen();
 
+# apply firmware previously copied to the node via SCP into /tmp/web/sideload_firmware.bin
+if($parms{button_apply_fw} and -f "/tmp/web/sideload_firmware.bin")
+{
+    system "mkdir -p /tmp/web/upload";
+    system "mv -f /tmp/web/sideload_firmware.bin /tmp/web/upload/file";
+    $parms{firmfile} = "arednmesh-sideload-sysupgrade.bin";
+    $parms{button_ul_fw}="Upload";
+    system("/usr/local/bin/uploadctlservices","upgrade");
+}
 
 # upload fw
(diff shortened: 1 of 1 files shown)


k1ky:
Now that's just "too cool"!  I've rarely have difficulty over marginal links, but this most certainly would be a great feature and would provide a good failsafe.

Orv:
What Tom said!   :-)

r1de:
awesome! now to hack that change into some of the persnickety nodes running 3.16, then I can finally bring them up to date!
Thx Darryl!

r1de:
If this gets done, `/tmp/web/` needs to be a permanent directory. Created at each boot.
It's not there unless you have already loaded the "admin" page and gets wiped out on reboots.

I fought with this many times trying to get some of the very marginal Santa Barbara nodes upgraded.

dman776:
By its very definition, /tmp is temporary since it is RAM and erased at boot time

Thanks,
Darryl

dman776:
Any thoughts on this?  It may be abandoned since it would involved scp based file transfers.

r1de:
I know /tmp is "temporary", couldn't there be a "mkdir /tmp/web" command somewhere in the init scripts? rc.local perhaps? (or even just move the uploaded files location to simply "/tmp")
I have had times I couldn't even load the GUI pages, but I could still scp the file and have it there, waiting, until I could manage to click the button...
I think this feature is an excellent addition to the "tool box", even if it does require the use of scp.
I have had to use it more than a few times already.

r1de:
To add one more thing,
This is one of those tools you may not even know you'll need, until you need it, and don't have it.
(again, just my $0.02 :smile: )

dman776:
Abandoning for now.

PR #196, opened 2018-09-22
aredn: Prepare 3.18.09.0RC1 for release

ae6xe:


config.mk
```diff
@@ -1,18 +1,18 @@
...
 # https://openwrt.org/docs/guide-developer/usebuildsytem (openwrt-lede merge)
-# example "MAKE_ARGS=-j33 V=s IGNORE_ERRORS=m BUILD_LOG=1"
+# example "MAKE_ARGS=-j9 V=s IGNORE_ERRORS=m BUILD_LOG=1"
 MAKE_ARGS=-j3
 
 # Where will the installed image find add-on Packages to download?
 # This URL must contain the packages from this build
 # downloading packages within the AREDN UI uses signatures 
-PRIVATE_BUILD_PACKAGES=http://downloads.arednmesh.org/snapshots/trunk
+PRIVATE_BUILD_PACKAGES=http://downloads.arednmesh.org/releases/3/18/3.18.09.0
```
dman776: remove the leading 0

files/www/cgi-bin/perlfunc.pm
```diff
@@ -1084,25 +1084,25 @@ sub hardware_info
     },
     '0xc302' => {
       'name'            => 'PicoStation2',
-      'comment'         => 'PicoStation2 (NOT M series) in testing',
```
dman776: the non-M series should not be supported

files/www/cgi-bin/perlfunc.pm
```diff
@@ -1084,25 +1084,25 @@ sub hardware_info
...
+      'comment'         => '',
+      'supported'       => '1',
       'maxpower'        => '16',
       'pwroffset'       => '4',
       'usechains'       => 0,
       'rfband'          => '2400',
     },
     '0xc3a2' => {
       'name'            => 'PicoStation2 HP',
-      'comment'         => 'PicoStation2 HP (NOT M series) in testing',
```
dman776: the non-M series should not be supported

dman776 (changes requested):

files/www/cgi-bin/perlfunc.pm
ae6xe: The picostation 2 has 8MB flash and 32Mb RAM.  I thought it did work.  "not M series" because it is the "M" series that is the issue --only have 4MB flash and no way to install.   (all the stuff we support is "M2").

dman776: It's never been on the SPM.  Only the PS M2

ae6xe: OK, let me take it out until we can confirm it works.

ae6xe:
Updates made.   Note the comment usage was inconsistent.  Many had no comment, some had a comment with same information as the name.    Unless the comment was different than the name and added more information,  it is now empty.

dman776 (approved):
Looks good

PR #217, opened 2018-10-04
feature: Provide firmware file advice for uploading firmware files

### LINKED ISSUE #206: Feature Request - Verify Node Board id before upgrading Firmware - match Firmware to appropriate Model for compatibility
Request that firmware compatibility to node make/model/board id type be checked before upgrading. If nothing more than a "warning" of possible incompatibility such as installing XM firmware on an XW device, mismatched Model i.e. installing Rocket Firmware on a Nanostation for example.

On several occasions even the most experienced user can accidentally install upgrade firmware on remote nodes, rendering them as "bricks", sometimes requiring a trip up a tower or a long trip to a remote site.

I can see situations where a total "lockout" for upgrading would not be preferable, but at least some sort of compatibility detection with a warning could avert imminent upgrade disaster.

dman776:
also reduces page size by removing unnecessary style info on option tags.

fixes #206

### DIFF
--- a/files/www/cgi-bin/admin
+++ b/files/www/cgi-bin/admin
@@ -120,11 +120,11 @@ if($mfg=~ /Ubiquiti/i)
 
 if($hardwaretype eq "nanostation-m")
 {
-    $hardwaretype="nano-m";
+    $hardwaretypev="nano-m";
 } elsif($hardwaretype eq "rb-952ui-5ac2nd") {
-    $hardwaretype= "rb-nor-flash-16M-ac" ;    # hAP AC Lite
-} elsif($hardwaretype =~ /rb-912uag-[25]hpnd/ ) {
-    $hardwaretype= "nand-large" ;    # Basebox 2/5
+    $hardwaretypev= "rb-nor-flash-16M-ac" ;    # hAP AC Lite
+} elsif($hardwaretype =~ /rb-912uag-[25]hpnd/i ) {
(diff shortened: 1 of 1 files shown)


wu2s (approved):
Approved

ae6xe:
I can test on several of the devices I have.   Any I should skip that you've already tested on?

dman776:
I’ve tested on Nanostation M2 with Firefox and Chrome.

ae6xe:
image

ae6xe:
The hap ac lite images (all the mikrotik images) aren't very nicely named.  There's nothing about the model # in the filename.  In this case the correct image is:
aredn-3.18.9.0-mikrotik-rb-nor-flash-16M-ac-sysupgrade.bin

The above error was given in trying to load the correct name.   The tp-link devices have similar issue.

ae6xe:
Tested successfully on hAP ac lite and basebox 5.

ae6xe:
issue on cpe210v2:
image

dman776:
What does /usr/local/bin/get_hardware_mfg return on this device?

ae6xe:
get_hardware_mfg, get_boardid, and get_model all return:
TP-Link CPE210 v2.0

ae6xe:
on cpe510 v1.1:
get_hardware_mfg = TP-Link
get_hardwaretype = cpe510
get_model = TP-Link CPE510 v1.1
get_boardid = TP-Link CPE510 v1.1

PR #224, opened 2018-10-16
bugfix: check to see if internet is reachable before loading resources

### LINKED ISSUE #223: Internet Connectivity Flags
There's been some discussion lately about various services in a node that require Internet connectivity in order to function. Perhaps it's time for a node to be aware internally if it has Internet connectivity and how? These flags would be used by all services requiring Internet connectivity and control whether they operate at all. This might fix some timeout issues being reported and overall provide a cleaner operating node.

* DNS
* WAN Gateway (Issue 200)
* Mesh Gateway (Issue 214)
* UI Delay (Issue 222)
* NTP (inform sysop that NTP souce must now be on net)
* Others ?

dman776:
Fixes #223

### DIFF
--- a/files/www/cgi-bin/perlfunc.pm
+++ b/files/www/cgi-bin/perlfunc.pm
@@ -1704,7 +1704,11 @@ sub is_online()
   my $online=0;
   if(get_default_gw() ne "none")
   {
-    $online=1;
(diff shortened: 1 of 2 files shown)


ae6xe (approved):
ping the leafjets URL/hostname directly?   But this is sufficient.

PR #225, opened 2018-10-17
feature: add a "firstboot" button on setup page

### LINKED ISSUE #162: Please add "Just flashed" reset from software
Can we have a "just flashed" reset option through the AREDN software for those resets that need to take place when nodes are hard to reach... like on a roof or a tower and you can't easily push the reset button for 15 seconds?

dman776:
fixes #162


ae6xe (approved):
Instead of "Are you sure?", how about "All config settings and add-on packages will be lost back to first boot state.  Continue?"     I tested this on a cpe510 v1.1 with success.

PR #226, opened 2018-10-19
bugfix: Check for existing hostnames on the network in DHCP reservations

### LINKED ISSUE #216: Avoid duplicate hostnames when reserving IP addresses
Generic hostnames can be created across the mesh network, e.g. "RPi", "RasPBX", "ipCam", etc.,   If duplicate, one device may not be reachable.    Check if an existing hostname already exists for reserving an IP address.

r1de:
fixes #216
*additional check for existing hostname when creating new DHCP lease.
checks hosts_olsr file if entered hostname already exists or not on the connected network(s).
outputs info saying such and gives advice to prefix hostname with callsign.

### DIFF
--- a/files/www/cgi-bin/ports
+++ b/files/www/cgi-bin/ports
@@ -309,6 +309,22 @@ foreach $val (@list)
                 last;
             }
         }
+        my $olsrFile = 0;
+        $olsrFile = 1 if -f "/var/run/hosts_olsr";
+        my $foundHost = 0;
+        if($olsrFile) {
+          open(my $hostFile, "<", "/var/run/hosts_olsr");
+          while(<$hostFile>) {
+            if($_ =~ /\t$host\t/) {
+              $foundHost = 1;
+              last;
+            }
+          }
+          close($hostFile);
+          push(@dhcp_err, "$val <font color='red'>Warning!</font> '$host' is already in use!<br>" .
+            "Please choose another hostname.<br>" .
+            "Prefixing the hostname with your callsign will <em>ensure</em> it is unique.") if $foundHost == 1;
+        }
     }
     else
     {
--- a/files/www/cgi-bin/ports
+++ b/files/www/cgi-bin/ports
@@ -315,7 +315,7 @@ foreach $val (@list)
         if($olsrFile) {
           open(my $hostFile, "<", "/var/run/hosts_olsr");
           while(<$hostFile>) {
-            if($_ =~ /\t$host\t/) {
+            if($_ =~ /\s$host\s/i) {
               $foundHost = 1;
               last;
             }
--- a/files/www/cgi-bin/ports
+++ b/files/www/cgi-bin/ports
@@ -323,7 +323,7 @@ foreach $val (@list)
           close($hostFile);
           push(@dhcp_err, "$val <font color='red'>Warning!</font> '$host' is already in use!<br>" .
             "Please choose another hostname.<br>" .
-            "Prefixing the hostname with your callsign will <em>ensure</em> it is unique.") if $foundHost == 1;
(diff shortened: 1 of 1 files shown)


ae6xe (changes requested):
The error message when displayed is chopped off.  it only shows the first line, "hostname `xyz` is already in use".     The other 2 lines aren't shown -- remove from code or find a way to show them.   (I'm using chrome.)  

Also, not sure why this code works with different combinations of upper and lower case letters, e.g. it still finds duplicate with "XYZ" and "xyz".   The line with "/\t$host\t/" may need explicit case insensitive match, "/\t$host\t/i" to be sure.    Maybe case insensitive mode has been turned on somewhere else somehow?

ae6xe:
image

dman776:
the default match is case-sensitive.   You certainly need the /xxxx/i   syntax to disable it.

ae6xe:
Also,  should probably use "/s" instead of "/t".   May not be guaranteed to always have a tab char.

r1de:
copy that...
I bet the text is getting truncated when the lease list is already large... (I only tried it on a test node with no other leases)

I'll try it on one line, but I'm not sure that'll do it... (or try to shorten the message)

r1de:
I wonder if it is just after that 5th/6th entry:
arednhostnametest02

I'll try to track that down...
Never say "that looks easy!" :)
(this is with firefox on linux)

r1de:
`/\s$host\s/i`
sorry for typo in the commit

r1de:
Tested the warning message on every browser I have in my house, chrome, firefox, *cough* IE, etc..
I've used tablet(s), phone(s), and PC(s).
I do not see the warning being truncated with any of them...

Is anyone else seeing the same behavior Joe is seeing?

Joe, can you check and make sure you have the latest code for this PR? I think you somehow have an older version of it.

ae6xe:
That was the original code submitted and results.  I'll retest tonight with current code.

r1de:
Ok, Thx. I'll double check I submitted the correct stuff (I am pretty sure I did).
It's just odd that you don't even have the "Warning!" part of the text. 

```
          push(@dhcp_err, "$val <font color='red'>Warning!</font> '$host' is already in use!<br>" .
            "Please choose another hostname.<br>" .
            "Prefixing the hostname with your callsign will <em>ensure</em> it is unique.") if $foundHost == 1;
```
(above code copied from here: https://github.com/aredn/aredn_ar71xx/pull/226/files )

dman776:
Change: "Prefixing the hostname with your callsign will <em>ensure</em> it is unique."
to: "Prefixing the hostname with your callsign will help prevent duplicates on the network."

ae6xe:
I'll inspect the code still on my test node to make sure there's not an
error in how I copied code down and installed.  Could a long hostname be a
factor, pushing the string outside of edges?

r1de:
> I'll inspect the code still on my test node to make sure there's not an error in how I copied code down and installed. Could a long hostname be a factor, pushing the string outside of edges?

I think it'll just word wrap it to fit into the "area". My long sentences have been getting wrapped, I'd imagine a long hostname would be treated the same...

and copy that, @dman776 that wording does sound better.

ae6xe:
I'm getting this with the latest.   
image

r1de:
Ouch! that's what I get for testing this on a node with no other reservations! :smile:
That's a bit crazy that they *all* get validated every single time...
I think I know what it is...

maybe I can even catch existing dupes now tho.

r1de:
@ae6xe try it now. It should not do that anymore.
I was trying to also catch any existing duplicate hostnames, but it was not working, I'm sure I am just doing it wrong.
This will only catch *newly added* hostnames that are duplicates...

ae6xe (approved):
tested on hAP ac lite.  worked as expected.

PR #235, opened 2018-10-22
Remove aimer script that was never completely production ready

dman776:


K6AH:
The Aimer routine was actually used quite effectively for a few months.  But I agree we should remove it.

dman776:
Yes.   It's not hooked into the web UI at all AND we now have the much more feature rich capability with realtime charts (per node)

ae6xe (approved):
Looks good.

PR #237, opened 2018-10-23
Add "Advanced config" page

### LINKED ISSUE #230: Add ability in Settings to define a location for locally hosted images and packages
Some mesh networks do not connect to the internet and are unable to 'refresh' in the Administration page to ease the ability to find and install newer firmware versions or install add-on packages.   Add a setting to enable definition of other locally hosted "download" web servers.    Instead of a  built in path of "http://downloads.arednmesh.org/releases/3/18/3.18.9.0/",  a group could clone these files and put locally on a web server at, e.g. "http://ae6xe-downloads.local.mesh/3.18.9.0".

dman776:
Adds settings for leafletjs, leafletcss, maptiles, and firmware download location
fixes #230

### DIFF
--- a/files/etc/arednsysupgrade.conf
+++ b/files/etc/arednsysupgrade.conf
@@ -17,6 +17,7 @@
 /etc/config.mesh/_setup.services.dmz
 /etc/config.mesh/_setup.services.nat
 /etc/config.mesh/vtun
+/etc/config.mesh/aredn
 /etc/config.router/_setup
 /etc/config.router/_setup.dhcp
 /etc/config.router/_setup.ports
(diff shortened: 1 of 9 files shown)


ae6xe:
not seeing the default settings for serverpath:
image

dman776:
what is in /etc/config.mesh/system for section aredn/option serverpath??

ae6xe:
On the live node:

config aredn
        option serverpath 'http://downloads.arednmesh.org/firmware/ubnt'

ae6xe:
after copying up the modified files, I probably need to "save" settings.  This isn't in /etc/config/system.

dman776:
correct.   if you are testing by copying the files in, the default is in config.mesh and you will need to SAVE SETTINGS to copy then to the live config in /etc/config

ae6xe (approved):
I tested on an hap ac lite.   changed firmware download to a local web server.  all worked as expected.

PR #240, opened 2018-10-25
bugfix: correct improper "wrong firmware" message on Rocket

dman776:
was not properly setting the hardwaretypev variable on some devices


ae6xe (approved):
looks good

PR #245, opened 2018-10-26
feature: allow upload of ssh keys while tunnels since keys

### LINKED ISSUE #244: Allow ssh keys to be uploaded while tunnel is enabled
Currently the Upload button for uploading ssh keys is greyed out when a tunnel is enabled.  Per Joe, it's not memory intensive compared to a sysupgrade.   So recommend not greying out the Upload button for ssh key upload.

dman776:
fixes #244


ae6xe (commented):
looks good.

ae6xe (approved):

PR #247, opened 2018-10-26
modify the advanced config layout

dman776:


ae6xe (approved):
add, "aredn:  modify the advanced config layout"  to the commit description.

Orv:
I think I know how this is supposed to work.  I modified the firmware path to http://w6bi-house-pc/aredn [after verifying that that URL is legitimate), saved it and rebooted the node.    But when I click on the Upload Firmware button it still opens the last place loaded (on my home computer).  

Am I doing something wrong?  This is with nightly build 379.

ae6xe:
When you "download" the firmware from this server path, click on the "download" button.   When you "upload" from your desktop/laptop computer, click the "upload" button.    How might we re-work the UI so this confusion doesn't hit others?     I suspect almost no one has used the "download" before, so we are all conditions to click "upload".  The "keep settings" option is also only aligned with the "download" row, so why would we think it applies to the "upload" row?...

Orv:
"Doh".    Yeah, it needs to be a bit more obvious.   Now that I got THAT figured out, I put the nightly build at the URL that's configure  (a .bin file), and hit the Refresh button.  I expected to see a .bin file listed.  Instead, I got this:

Downloading firmware list from http://w6bi-house-pc/aredn...
Downloading 'http://w6bi-house-pc/aredn/firmware.nanostation-m-xw.list'
Connecting to 10.95.109.109:80
HTTP error 404

Where does that list come from?    And is that the expected behavior?

ae6xe:
copy down from "http://downloads.arednmesh.org/firmware/ubnt/"  the  *.list files. These files list the firmware options available by model -- a row for each image version to select from.    To add the nightly build, add a row or edit existing to the appropriate .list file to include this selection and accordingly copy down the image to your local web server.     The list has an md5 signature for the image.  This can be generated with "md5sum \<image\>"

dman776:
Change it in advanced config.
Go to admin page ( no reboot)
Click REFRESH under the firmware section
Pick 3.18.9.0

Thanks,
Darryl

Orv:
I did download the *.list files.   Now I don't get an error when I refresh - but what about nightly builds? That's what I was primarily interested in at this point.  (Especially since the feature only exists in the nightly builds - I'd only be installing 3.18.9.0 if I was reverting the node back).

Orv

ae6xe:
manually edit the .list files to add a row and define the nightly build images.   The UI selection is defined in the respective .list file.

Orv:
You did say that, didn't you?   It didn't sink in.  I'll give it shot.

Orv

Orv:
I can do that.   But how is the hex number derived?  and what is the 
meaning of 'all' on the end?

Example:

73ba46bbd0b2e70588d870279596b4cf 
aredn-3.18.9.0-ubnt-nano-m-xw-sysupgrade.bin all

dman776:
Md5sum of the file

Thanks,
Darryl

Orv:
I suspect as much - thanks.

Orv

PR #250, opened 2018-10-27
bugfix: correct poe functions in adv cfg

dman776:
cleanup shellcheck errors


ae6xe (commented):
Also, avoid another shell:

from - export pin=$(echo "gpio${pinnum}")
to     - export pin="gpio${pinnum}"

ae6xe (approved):
looks good.

PR #252, opened 2018-10-29
enhancement: Custom DNSMasq options

### LINKED ISSUE #251: enhancement: DNSMasq custom options
Just like custom firewall rules, but for DNSMasq.
Allow for custom options, such as "mx-host". :)

r1de:
Allows additional dnsmasq options that will be saved across upgrades.
Fixes #251 
(once I saw how stupidly simple this was, I couldn't let it go by. :smile: )

### DIFF
--- a/files/etc/arednsysupgrade.conf
+++ b/files/etc/arednsysupgrade.conf
@@ -25,6 +25,7 @@
 /etc/dropbear/dropbear_rsa_host_key
 /etc/dropbear/authorized_keys
 /etc/local/mesh-firewall/59-custom-rules
+/etc/local/dnsmasq/dnsmasq-custom.conf
 /etc/firewall.user
 /etc/group
 /etc/hosts
--- a/files/etc/dnsmasq.conf
+++ b/files/etc/dnsmasq.conf
@@ -14,6 +14,9 @@ resolv-file=/tmp/resolv.conf.auto
 # include olsr nameservice
 addn-hosts=/var/run/hosts_olsr
 
+# additional custom options
+conf-file=/etc/local/dnsmasq/dnsmasq-custom.conf
+
 dhcp-authoritative
 dhcp-leasefile=/tmp/dhcp.leases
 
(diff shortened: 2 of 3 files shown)


ae6xe:
Eric,  Maybe there is a way to hook this into the advance config screen?  We have a boolean and string capability on this new UI page, what about a text file option?   This would potentially extend the advanced config to more than dnsmasq user config options.  (doesn't necessarily need to be included in this commit.)

r1de:
There probably is, and with those functions in the adv config page it shouldn't be hard.
I had already thought of that, but I am just getting over being sick as heck and I've only begun to be able to think again. :smile:

dman776:
we need a GUI way to do this if we are going to implement it.

r1de:
Ok. I'll look at that tonight

dman776:
I think we should list explicit options on the adv config (AC) page for each option that we want to expose, instead of here's a block of text.  ie. lets add an AC entry for MX record, one for TFTPServer, etc.   Then, have a shellscript that simply builds the option file from the uci values.  The shellscript would be called from a postcallback function on the AC page.

dman776:
something like this:
push @setting, {
  key => "aredn.\@dnsmasq[0].mx", 
  type => "string", 
  desc => "Specifies ..........", 
  default => "1", 
  postcallback => "buildDNSMasqOptionFile()"
};

r1de:
There might be several "mx-host" options tho.
Could it handle that?
I have one node that has 3 "mx-host" option lines currently, all pointing at different smtp servers.

ae6xe:
This aredn config setting could be a string, something like "mx-host1 ; mx-host2 ; mx-host3 ; ..."   The under-the-hood script could convert this to the multi-line conf settings?

r1de:
I'll give it a shot, but it sure seems like overkill just to edit a simple text file. :smile:

r1de:
Closing this.
I am still working on it and will submit a new one shortly but based on the newer code.
This is too far behind.

PR #253, opened 2018-10-31
enhancement: kernel build options, additions to top/busybox, modules

r1de:
Added some kernel build options trying to conserve more RAM.
I would say at least a couple 100K more free RAM on average (perhaps more).
I have been running these options on my custom testing builds for a few weeks now.

disable CRASHLOG interface, we can't really see it anyway
CONFIG_KERNEL_CRASHLOG=n

remove process core dump support (see above reasoning)
CONFIG_KERNEL_ELF_CORE=n

there is no "swap" so why even compile it in?
CONFIG_KERNEL_SWAP=n

The following 6 options probably make the most difference:

remove debug symbols
CONFIG_KERNEL_KALLSYMS=n

remove debug info
CONFIG_KERNEL_DEBUG_INFO=n

even more stuff from the kernel we don't need is gone!
CONFIG_STRIP_KERNEL_EXPORTS=y

strip the libraries of unnessesary functions!
CONFIG_USE_MKLIBS=y

compress /tmp (seems to be minimal overhead to this)
CONFIG_PROCD_ZRAM_TMPFS=y

These are more of remove what we can't use/don't need

have you ever used this key stroke? I haven't. :)
CONFIG_KERNEL_MAGIC_SYSRQ=n

This is not used with AREDN.
CONFIG_PACKAGE_MAC80211_MESH=n

All above options taken from: https://openwrt.org/docs/guide-user/additional-software/saving-space#modifying_build_configuration_variables

-------------

I added these 2 in for good measure, they aren't needed and never really were, this is minimal saved space:
CONFIG_BUSYBOX_DEFAULT_MKSWAP=n
CONFIG_BUSYBOX_CONFIG_MKSWAP=n

busybox "top" sucks, this makes it suck slightly less and act more like the normal "top" command.
CONFIG_BUSYBOX_CONFIG_FEATURE_TOP_INTERACTIVE=y

This other USB stuff was now made into modules to be installed if needed.
CONFIG_PACKAGE_kmod-usb3=m
CONFIG_PACKAGE_block-mount=m
CONFIG_PACKAGE_kmod-usb-storage=m
CONFIG_PACKAGE_kmod-usb-storage-extras=m
CONFIG_PACKAGE_kmod-usb-storage-uas=m

Additional filesystem modules are not really needed 99.9% of the time.
Same with the Ext2FS utilities.
CONFIG_PACKAGE_kmod-fs-vfat=m
CONFIG_PACKAGE_kmod-fs-ntfs=m
CONFIG_PACKAGE_kmod-fs-ext4=m
CONFIG_PACKAGE_e2fsprogs=m

These 2 are in there, but I am not sure are really doing anything.
IIRC, they told the kernel to hold less RAM for log files(?) which we barely have anyways.
I can't find my notes of course...
CONFIG_LOG_BUF_SHIFT=16
CONFIG_LOG_CPU_MAX_BUF_SHIFT=16

The following 3 options can be added to give some additional info for debugging/dealing with some persnickty issues maybe. :)
CONFIG_BUSYBOX_CONFIG_IOSTAT=y
CONFIG_BUSYBOX_CONFIG_RENICE=y
CONFIG_BUSYBOX_CONFIG_FEATURE_PS_LONG=y


ae6xe (commented):
The gpsd and libgps options can be removed and don't explicitly need to be defined.    These are already building because they are a dependency of an olsrd plugin that requires them.   I generally have tried to avoid explicitly defining something to build that is already being built as a dependency, to keep the config definitions more manageable.

r1de:
I did not catch that...

done.

ae6xe:
MAC80211 MESH can be removed.  It is a different 802.11 mode than the adhoc mode we are using.  It is not used in AREDN.   No issues.  Can change the comment to be simply unused in AREDN.

r1de:
> Can change the comment to be simply unused in AREDN.



done...

ae6xe:
CONFIG_TARGET_SQUASHFS_FRAGMENT_SIZE=2

This saves 256 Bytes, correct?  I'm concerned lowering the FS cache could potential have more of a performance impact or the tradeoff is not worth it.    But I'm not sure how to quantify this to really compare the trade-offs.   given the negligible savings, I'd vote don't do this one for now.

r1de:
I haven't seen an issue yet, but I do agree, we shouldn't push it too much.

done...

ae6xe:
update the commit  comments to be consistent with the changes.  I'm good with these options.

r1de:
done. Thanks Joe!

r1de:
Did you want to remove MAC80211 MESH from the options Joe? I may have misread that comment.

ae6xe:
Sorry, leave the config option in that removes 80211 mesh  from the kernel.    It can be removed from the kernel as AREDN doesn't use this protocol.

dman776:
I will build this later today and load on a few devices.

r1de:
Edited initial commit description, hopefully I did that correctly and didn't mess up anything.

dman776:
On hAP AC Lite between build 385 and this PR:
saved 520KB total memory
reduced used memory by 3624KB
increased free memory by 4144KB

No negative effects noted.

dman776 (approved):
looks good

Issue #241, opened 2018-10-26
Turn power on/off on USB port if available on a device


ae6xe:
Devices with USB ports might be used to power a local RasPi or other devices.   Enable an advanced setting to turn USB power on/off.  

to manually do this on the hAP AC lite device:
echo 45 > /sys/class/gpio/export           # if /sys/class/gpio//gpio45 does not exist
echo 1 > /sys/class/gpio/gpio45/value  # to turn on
echo 0 > /sys/class/gpio/gpio45/value  # to turn off
Be sure to stay within the hAP AC lite specs of 1A on the USB port.

Orv:
Is this persistent, Joe?

ae6xe:
No, not persistent.

dman776:
Once this feature is implemented it will be persistent 

Thanks,
Darryl

AC0WN:
Tested and works great.  Looking forward to the implementation.  :)

Thanks,
julie

PR #239, opened 2018-10-24
Add PoE Passthru power support

dman776:
fixes #229

PR #255, opened 2018-11-02
feature: add USB passthrough support to advancedconfig page

dman776:


dman776:
Fixes #239 
Fixes #241

ae6xe (commented):
check shellcheck issues.    the "export pinnum=`cat ... `"   should be "pinnum=$(cat ...)"

ae6xe (approved):
looks good.

PR #265, opened 2018-11-10
bugfix: litebeam M5 is single chain device

ae6xe:


dman776 (approved):
Datasheet shows a single chain (for the non-AC models).   This looks good.

PR #267, opened 2018-11-14
bugfix: suppress uhttp uci error on mesh page with no description

dman776:


ae6xe (approved):
That will work, looks good.

PR #269, opened 2018-11-15
bugfix: AREDN standard port vlan function on Ubiquiti NS XW devices

ae6xe:
The NanoStation M2 and M5 XW switch configuraiton had LAN
on primary port (untagged) and DtDLink and WAN (vlan 1 and 2)
on secondary port. This change enables untagged, vlan 1,
and vlan 2 on primary port to function.  This configuration
also functions on the secondary port.  However, devices on
the same network can not communicate with one another across
the two ports. Although, access works from the greater mesh
and the node to devices on both ports on the same network.
When using one port the behavior is now consitent with all
other AREDN devices. The use of both ports offers some
benefit, but with the above limitations.


dman776 (approved):
works fine

PR #270, opened 2018-11-15
bugfix: Found another place where uhttpd was tossing a uci error. Fixed

r1de:
Also removed the description block from the mesh page if the description has not been set.
(there was a small grey area before, even with no description set)


ae6xe (approved):
Looks good.

PR #271, opened 2018-11-17
bugfix: correct initial values for poe and usb passthrough

dman776:
if values are not yet in /etc/config/aredn, then:
- pull the default values from board.json
- set value in /etc/config/aredn


ae6xe (approved):
looks good.

PR #272, opened 2018-11-17
aredn: Add definition for Mikrotik LHG-5HPnD-XL

ae6xe:


dman776 (approved):
looks good.   Ray tested this.

PR #275, opened 2018-11-22
aredn: enable nano-m-xw switch to be a dumb switch

ae6xe:
Config the NanoStation M2/M5 XW devices to be fully
functional with a standard aredn port definition.
The vlans are not configured in the switch, rather
acting like a dumb swtich, enabling tagged and untagged
packets to exist on the same switch port at the same time.
This commit enables the ports on the switch to directly
send all traffic to all other ports when not configured
directly in vlan mode.


dman776 (approved):
looks good

PR #276, opened 2018-11-22
uhttpd: serve precompressed files

ae6xe:
The aredn project benefits from precompressed files given
mesh nodes are typically long distant RF links and UI
is accessed remotely.  uhttpd will look for the filename
requested first, then if not found, look for filename.gz
and update content header passed back as appropriate.

https://patchwork.ozlabs.org/patch/514918/

### DIFF
--- a/patches/712-uhhtpd-compress.patch
+++ b/patches/712-uhhtpd-compress.patch
@@ -1,0 +1,131 @@
+Index: openwrt/package/network/services/uhttpd/patches/001-uhttpd-compress.patch
+===================================================================
+--- /dev/null
++++ openwrt/package/network/services/uhttpd/patches/001-uhttpd-compress.patch
+@@ -0,0 +1,113 @@
++--- a/file.c
+++++ b/file.c
++@@ -139,6 +139,7 @@
++ 	int docroot_len = strlen(docroot);
++ 	char *pathptr = NULL;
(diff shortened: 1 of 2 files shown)


dman776 (approved):
tested successfully on local hAP AC Lite

Orv:
So this is intended for firmware uploads?  Any other uses?

ae6xe:
This is mostly in preparation for a new UI.  We're evaluating an "angular" front end framework for the browser side.   Although, there are some files used by the charts that could be compressed now and lower the up-front download time to the browser.  The files or code in scope is essentially the processing the client web browser needs to do.  We want to offload processing to the browser client off of the mesh node where possible.  This change enables the web server, if it can't find a file, it will then fallback and look for a (much smaller) .gz file and send instead.

dman776:
No.  It is to reduce the bandwidth required to deliver resource files like JavaScript libs and css files.

Thanks,
Darryl

dman776:
For example, after compressing a few files, the charts page goes from 700KB of bandwidth to around 400KB

Thanks,
Darryl

stokito:
FYI: I trying to convince openwrt devs to include the patch into master
http://lists.infradead.org/pipermail/openwrt-devel/2020-June/023870.html

PR #278, opened 2018-11-23
feature: reduce bandwidth by compressing resource files

dman776:
Gzip most large resource files to reduce the bandwidth required to initially send these across the mesh.


ae6xe (approved):
looks good.

PR #282, opened 2018-11-24
Revert: uhttpd precompression

ae6xe:
Some issues still linger, will pull back into development and resubmit later.


dman776 (approved):
Ok to revert

PR #283, opened 2018-11-26
bugfix: usb_passthrough script missing execute bit (+x)

dman776:
/usr/local/bin/usb_passthrough did not have it's execute bit set on the filesystem and could not execute.


ae6xe (approved):
looks good.

PR #285, opened 2018-11-28
bugfix: POE passthough fails on NS XM devices

ae6xe:
the Nanostation XM devices require setting the gpio
port direction unlike some other devices


dman776:
have you tested this on a mikrotik?   If I recall, they didn't like the direction values.

ae6xe:
the code checks if 'direction' exists, if so then it echos 'out'. I have tested on an hAP ace lite.

dman776 (approved):
looks good

PR #286, opened 2018-11-28
aredn: stop cron prior to starting sysupgrade

ae6xe:


dman776 (approved):
looks good.

Orv:
What's the rationale behind this?  (Just curious).

dman776:
This should eliminate the "failed" sysupgrade issue where after a load attempt, the node is still at the same version as before.

ae6xe:
We found that when preparing to do the upgrade process, and programs and RAM were freed up, that cron was restarted by procd (or stayed running) and was continuing to starting up new processes in the opposite direction causing havoc on 32MB RAM devices.

PR #287, opened 2018-11-28
aredn: hAP ac lite enable 5GHz LAN Access Point

### LINKED ISSUE #215: hAP lite 5Ghz enable dual band capability

ae6xe:
enable ability to bridge LAN physical ports with wireless
LAN Access Point capability. On dual band hAP ac lite,
used in parallel with mesh RF on 2GHz.  Enables future
features on single band devices to turn off mesh RF and
repurposed for mesh LAN Access Point.

closes: #215


ae6xe:
I've tested this on the following devices:
-hap ac lite
- NS M2 XM
- NS M5 XW
- cpe510v1.1

files/www/help.html
```diff
@@ -407,11 +405,10 @@
 covered here.
 </p>
 <p>
-The other option in the WAN box is the <strong>Mesh Gateway</strong>.
-This is an advanced configuration option.<br>
-When a node has internet access from either the WAN or LAN, that access is
-available to the node itself and to any computer connected to the LAN port.
-When the Mesh Gateway is enabled this node will gate(route) traffic from the
+The <strong>Allow others to use my WAN</strong> is an Advance WAN option.
```
dman776: change Advance to Advanced

dman776 (changes requested):

files/www/help.html
ae6xe: Corrected.

dman776 (approved):
tested on my hapaclite and rocket m2

AC0WN:
Build:  aredn-491-a62fe95-mikrotik-rb-nor-flash-16M-ac-sysupgrade.bin

If your SSID has one or more SPACES in the name the 5 GHz radio does not broadcast when the 5G settings are saved and the router rebooted.  We checked this with a WiFi scanner on several different channels.

As soon as you replace the SPACES in the SSID with UNDERLINES <_> the radio works great.

ae6xe:
good catch.   Spaces are allowed.  Will submit a change shortly to allow the spaces to correctly save and pass though the config settings.

ae6xe:
See submitted change:   https://github.com/aredn/aredn_ar71xx/pull/293

PR #289, opened 2018-11-29
feature: provide more info on upload disable with tunnels

dman776:
Provider additional user info on why UPLOAD buttons are disabled when tunnels are enabled.


ae6xe (approved):
this will help.  looks good.

PR #292, opened 2018-11-30
bugfix: usb passthrough not initializing after reboot properly

### LINKED ISSUE #291: USB power on mikrotik hap lite 952ui-5ac2nd
Can the team find a way to keep the USB power on after you power down and up?  It must be manually saved to restore the power each time however, the PoE works from startup.

dman776:
fixes #291


ae6xe (approved):
Yes.  Needed change.

PR #293, opened 2018-11-30
aredn: harden hAP ac lite 5Ghz Lan Access Point

ae6xe:
enable passwords with virtually any character,  enable SSIDs
with virtually any character.  ensure ap is always
encrypted to give operator control of client access and
license compliance.  Remove (week) WEP encryption option.


ae6xe:
The commit turns the ssid and password into hex encoded string and stores in aredn _setup.   When the wireless config settings are created, the hex string is then converted back to the original characters.  OpenWrt SSID and password could be virtually in characters in the wireless config file.

dman776 (approved):
good changes.

Issue #219, opened 2018-10-05
Add support for TP-Link 610


kylehamilton:
Hi I'm here from the forums I have a TP-Link CPE 610 (v1.0). I've included the requested hardware information if there's anything else you need just let me know.

Device Name:
CPE610
Device Model:
CPE610 v1.0
Firmware Version:
2.1.13 Build 20180327 Rel. 54027 (5553)

# cat /proc/cpuinfo
system type             : Atheros AR934x
processor               : 0
cpu model               : MIPS 74Kc V4.12
BogoMIPS                : 279.55
wait instruction        : yes
microsecond timers      : yes
tlb_entries             : 32
extra interrupt vector  : yes
hardware watchpoint     : yes, count: 4, address/irw mask: [0x0000, 0x0098, 0x04                                                                                                                                                             30, 0x0118]
ASEs implemented        : mips16 dsp
shadow register sets    : 1
core                    : 0
VCED exceptions         : not available
VCEI exceptions         : not available

# dmesg
Linux version 2.6.31 (root@ubuntu) (gcc version 4.3.3 (GCC) ) #1 PREEMPT Tue Mar 27 14:45:22 CST 2018
flash_size passed from bootloader = 0
CPU revision is: 0001974c (MIPS 74Kc)
ath_sys_frequency: cpu srif ddr srif cpu 560 ddr 450 ahb 225
Determined physical RAM map:
 memory: 02000000 @ 00000000 (usable)
User-defined physical RAM map:
 memory: 04000000 @ 00000000 (usable)
Zone PFN ranges:
  Normal   0x00000000 -> 0x00004000
Movable zone start PFN for each node
early_node_map[1] active PFN ranges
    0: 0x00000000 -> 0x00004000
On node 0 totalpages: 16384
free_area_init_node: node 0, pgdat 802429b0, node_mem_map 81000000
  Normal zone: 128 pages used for memmap
  Normal zone: 0 pages reserved
  Normal zone: 16256 pages, LIFO batch:3
Built 1 zonelists in Zone order, mobility grouping on.  Total pages: 16256
Kernel command line: console=ttyS0,115200 root=31:04 rootfstype=squashfs init=/init mtdparts=ath-nor0:128k(u-boot),64k(pation-table),64k(product-info),1536k(kernel),6144k(rootfs),192k(config),64k(ART) mem=64M
PID hash table entries: 256 (order: 8, 1024 bytes)
[...]

PR #295, opened 2018-12-02
ar71xx: Add Support for the TP-LINK CPE610 V1.0 Device

apcameron:
Installation:
 Flash factory image through stock firmware WEB UI
 or through TFTP
 To get to TFTP recovery just hold reset button while powering on for
 around 4-5 seconds and release.
 Rename factory image to recovery.bin
 Stock TFTP server IP:192.168.0.100
 Stock device TFTP adress:192.168.0.254

This Addresses Issue #219


ae6xe (approved):
looks good.

PR #297, opened 2018-12-03
aredn: Standardize Ethernet Port vlans for UBNT XM devices

ae6xe:
The secondary port on Ubnt XM devices was non-functional.
Enabled bridge type in network config to utilize both ports.


ae6xe:
I've tested on the following devices (new install and sysupgrade):
NSM2 XM
NSM5 XW
CPE201v2.0
hAP ac lite

KG6WXC has tested on:
NSM2 XM
NSM5 XM

dman776 (approved):
looks ok

PR #299, opened 2018-12-04
aredn: clarify distance setting in help

### LINKED ISSUE #279: Distance to FARTHEST Neighbor
I find this entry on the Node Setup page just a little confusing, and it has taken me a year or two to fully understand exactly what the entry does.
The "help" file, for those of us who actually do read it, does not in my view, explain just what is required and how it works. 
This entry is particularly important with WIFI Networks where your home Node may see more than one remote Node. 
So am I right in thinking that we should set this entry for the closest Node that we can see from our home node. So that all packets go to this Node only , and then  they would be forwarded on from there.? 
I have been making these adjustments on our Wellington WIFI Network, and it does "appear" to speed up the responses from the distant Nodes further down the line.
So can we change **"Distance to FARTHEST Neighbor"** to better reflect exactly what is required here.  Perhaps **"Distance to FIRST Neighbor"**.
Gordon

ae6xe:
closes #279

### DIFF
--- a/files/www/help.html
+++ b/files/www/help.html
@@ -367,10 +367,19 @@ in this case the max level will change when you save the settings and will be
 capped at the max level supported by the hardware for that frequency.
 </p>
 <p>
-The <strong>Distance</strong> setting adjusts the packet retry timer
-to account for stations that are very far away, presumably about 300 meters or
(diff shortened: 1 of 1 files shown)


dman776 (approved):
Good change

PR #302, opened 2018-12-08
aredn: add support for Mikrotik LDF-5nD

ae6xe:
tested new install and sysupgrade with keep settings.  All functionality appears to be working.  Did not test tunnels or other add-on packages.


dman776 (approved):
everything looks in order.  No device to test locally.

PR #305, opened 2018-12-11
bugfix: aredn: Advanced config settings not persistent though reboot

### LINKED ISSUE #303: Advanced Configuration settings not persistent through reboot
In nightly builds 522 and 529, the Advanced Configuration Settings '"aredn.@poe[0].passthrough" and "aredn.@usb[0].passthrough" don't persist through a reboot:

-  Uncheck each box
- Click Save Setting for each configuration parameter
- reboot node

Upon reboot, both boxes are checked again.   (haven't confirmed if passthrough is actually on though).
(confirmed behavior on two different Mikrotik routers).

ae6xe:
fixes #303


dman776:
Looks good.  Thanks Joe

Issue #155, opened 2018-08-24
Feature Request - Disable RF deck option in user interface


k1ky:
When operating in congested areas with multiple nodes connected via DTD in a close environment, it is desirable to disable the RF deck in order to reduce congestion on the network.  Leaving the "option" to re-enable RF when needed.  Alternative would be to set a different SSID and reduce power,  but would rather get rid of the extra RF when not needed and have the node available for service.  This can be handy when extra nodes are serving as Tunnel Clients and Servers to further reduce the workload.

dman776:
Fixed in PR#307

PR #307, opened 2018-12-12
aredn: Mesh RF enable/disable

ae6xe:
add feature to turn off Mesh RF if not in use.  When off,
section to repurpose wireless card to LAN bridged Access
Point becomes visiable.

### DIFF
--- a/files/etc/config.mesh/_setup
+++ b/files/etc/config.mesh/_setup
@@ -8,11 +8,12 @@ wifi_channel = 1
 wifi_chanbw = 20
 wifi_distance = 0
 wifi_country = 00
+wifi_enable = 1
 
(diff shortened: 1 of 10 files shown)


dman776 (approved):
tested ok on hapAClite

k1ky:
Isn't this the same feature enhancement requested in my #155?

ae6xe:
Tom, yes.   Once the Mesh RF checkbox is saved to disable, then another section shows a "LAN Access Point" option and settings.   You can then choose to enable the wireless to be ether mesh RF or a mesh LAN Access Point (but not both at same time, unless it is the hAP ac lite with 2 wireless adapters).

PR #308, opened 2018-12-12
bugfix: advanced config usb passthru errors 

dman776:
erros when aredn.usb uci section is not available


ae6xe (approved):
looks good, is needed.

PR #309, opened 2018-12-12
bugfix: remove SNR from signal archive page

### LINKED ISSUE #242: Remove/Activate SNR Display on Archive Signal Archive Page
Is there a need or reason to have the non-functioning Realtime SNR feature displayed on the Signal Archive Page?  If not - I suggest/request that it be removed from the header of the page or make it match the live value in real-time (preferred?)

dman776:
fixes #242


ae6xe (approved):
looks good.

PR #312, opened 2018-12-18
aredn: hap ac lite add 2G/5G lan ap options

ae6xe:
Add option for hap ac lite to select which band to
use for LAN AP option, 2GHz or 5GHz.  Also, ensure
all wireless cards are defined when disabled to
prevent default wireless config options.


dman776 (approved):
Looks ok.

PR #318, opened 2018-12-22
bugfix: aredn sysinfo.json fails with mesh rf off

### LINKED ISSUE #317: sysinfo.json File Becomes Unreachable in Non-Mesh RF Mode
When a node is in non-Mesh RF mode (no AP or with AP), the node throws an error when queried for the sysinfo.json file from /cgi-bin/sysinfo.json. This is the error

**Bad Gateway**
The process did not produce any response

This behavior has been seen on a hAP Lite running Nightly 556 and 3 different AirRouters running nightly 566. As soon as the node is back to Mesh on RF, sysinfo.json is available again.

ae6xe:
added json meshrf->status->on/off
fixes:  #317


dman776 (approved):
looks good.

PR #327, opened 2019-01-21
Some LUA Code Clean Up, API

raysuelzer:
This PR places the lua functions contained in `aredn/info.lua`, `aredn/uci.lua` into a "lua table" so that they do not pollute the global namespace and make it apparent which LUA module contains a specific function. 

It also adds @dman776 's API code, with some minor adjustments and bug fixes to 'cgi-bin'.


dman776 (approved):
Changes look reasonable.   Approved.

PR #339, opened 2019-01-26
aredn: olsrd: add olsrd watchdog

### LINKED ISSUE #338: Add olsr watchdog
If olsr locks up or blocks for any reason, the node would not be reachable over the mesh.  Currently, olsr is only restarted if the process no longer exists.  This request is to add a safety net to restart olsr if it is still running, but not processing events.   If such an unexpected event happens, olsr would be restarted, and mesh nodes at hard to access tower sites would still be reachable,

ae6xe:
Adding native olsrd watchdog on top of procd
ensures mesh nodes are always accessable.

closes #338

### DIFF
--- a/configs/common.config
+++ b/configs/common.config
@@ -63,6 +63,7 @@ CONFIG_PACKAGE_olsrd-mod-nameservice=y
 CONFIG_PACKAGE_olsrd-mod-txtinfo=y
 CONFIG_PACKAGE_olsrd-mod-jsoninfo=y
 CONFIG_PACKAGE_olsrd-mod-dot-draw=y
+CONFIG_PACKAGE_olsrd-mod-watchdog=y
 CONFIG_PACKAGE_olsrd-mod-secure=y
 CONFIG_PACKAGE_perlbase-essential=y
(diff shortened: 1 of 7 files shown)


dman776 (commented):
Is the 5sec interval too quick.   Seems like 60 would be sufficient

ae6xe:
The check to do the olsr restart is at 21 seconds intervals, giving olsr multiple attempts to touch this watchdog file.  A total time to recover node would be about ~50 seconds.  There can be timing issues when olsr is starting up, and to be conservative, we give olsr multiple opportunities to tickle the watchdog file.    If we changed it to,e .g. 60 seconds, then we'd also want to then only trigger a reset after 3+ minutes, which translates to a recovery time of 3.5+ minutes.    

In argument of the current value, the processing by olsr is relatively trivial -- it is touching a file in RAM.   I'd propose ether leave it as is, or a middle ground of say 20 sec tickle and 70 sec reset threshold, total recovery time about 100 seconds.

ae6xe:
minor update to touch the log file at start so the support download doesn't have error when file doesn't exist.

dman776 (approved):
Looks good

PR #342, opened 2019-01-29
feature: add sounds to realtime signal page

### LINKED ISSUE #341: Add audible indication to SNR screen
AirOS has a tool whereby a beep is periodically issued from the UI when reporting SNR - it's helpful when you are aligning an antenna and can't sit at a computer, but it's close enough to listen to.

Something like this would be handy within the AREDN firmware - beep every time the SNR screen refreshes (when on a specific node's signal). The beep would increase in frequency when the SNR improves and conversely lower in frequency when the SNR decreases.  (And maybe be silent below a given usable SNR?)

It could default to be always on, but more preferably have the ability to be toggled on/off.

dman776:
closes #341

### DIFF
--- a/files/www/cgi-bin/signal
+++ b/files/www/cgi-bin/signal
@@ -155,6 +155,7 @@ $header = <<EOF;
           dps[0].push(result[0][0]);
           chart.render();
           var snr = (result[0][0].y[1] * -1) - (result[0][0].y[0] * -1);
+          toneFreq(snr);
           \$('#snr').html(snr);
         });
       };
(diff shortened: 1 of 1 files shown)


ae6xe:
tested on 3 different nodes and tower sites.  It's very cool and works great with low to high SNR.  A -92dBm signal has a relatively low base signal, which booms on my subwoofer :) .

ae6xe (approved):
I like it.

Orv:
Thanks, guys!

PR #347, opened 2019-01-30
feature: add pitch control to signal tone

dman776:


K6AH (dismissed):
I recognize it's a little late for this, but it may make sense to add an "additive" component to this that wold set the lower limit of the tone.  Some computers aren't going to be able to produce low tones at any significant volume to be heard more than a few feet.

dman776:
What would you want the limit to be?   What SNR value?

Thanks,
Darryl

K6AH:
The lowest tone it should produce should be defined by a value (or slider)
entered.  I suppose you could just ask for it in Hz.   If 250Hz was
entered, then you would simply add 250 to the result of you calculation.  A
zero SNR would then produce a 250Hz tone.

Andre

dman776:
That’s exactly what this PR does.   The slider allows the user to change the pitch (ie.  Adds from 5 to 100 to the existing SNR, based on the “pitch” slider)

dman776:
correction.   I don't "ADD" the slider value to the SNR.   The code multiplies it to make a more pronounced tone change.

dman776:
Any further comments?

K6AH:
I believe I'm just overthinking it.  Let's go with what you've done and see
if there are operational concerns from users.

Andre

PR #353, opened 2019-02-03
feature: add volume control to signal tone

### LINKED ISSUE #348: Add Volume Control to Audible SNR Screen
The Audible SNR now coupled with the variable pitch control is extremely useful.  I had occasion to use it today with our man on the tower and me at the controls in the equipment building at a repeater site.  It became quickly apparent that there is a need for volume control that is easy to access when changing the pitch while one hand is busy with PTT on the radio and another on the computer keyboard.  Since we are already there in the SNR control panel it would be very handy to be able to adjust the volume from the same area instead of having to hunt around on the PC to change the volume level (which can be quite tricky on some machines with one hand .  This would make changes "on the fly" very easy.

dman776:
fixes #348

### DIFF
--- a/files/www/cgi-bin/signal
+++ b/files/www/cgi-bin/signal
@@ -183,6 +183,8 @@ $header = <<EOF;
     function toneFreq(snr) {
       var p = document.getElementById("tonePitch").value;
       oscillator.frequency.value = snr * p;  
+      var v = document.getElementById("toneVol").value;
+      gainNode.gain.value = v;
     }
 
     function toneOn() { 
@@ -191,7 +193,7 @@ $header = <<EOF;
       oscillator.connect(gainNode);
       oscillator.type = 'sine';
       gainNode.connect(audioCtx.destination);
-      gainNode.gain.value = 1; 
+      gainNode.gain.value = .5; 
       oscillator.start();
       document.getElementById("toneOff").disabled = false;
       document.getElementById("toneOn").disabled = true;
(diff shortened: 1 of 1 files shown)


ae6xe (approved):
looks good.

k1ky:
Need more range with the volume control - seems to hit "maximum" and goes into clipping after around 10% volume level (after 2nd click on the left).  Need the ability to attenuate a lot more.  Is this the proper place for this comment or do we need a new issue? N-678 + N683 on Rocket M5 XM, PBEM5-400, PBEM5-300, NSM2

dman776:
The gain is set to maximum on the slider.   The rest is up to you sound card settings.

k1ky:
It is clearly overdriven to the point of distortion above 10% setting (2nd click) irregardless of soundcard settings. Comparing to the level of previous build (before volume control) which has a clean signal.  In other words, it needs to be "padded" down quite a bit in order to give a good range to decrease the audio level. Need much more attenuation "range".

gelmce:
I just tried it out on a desktop with typically low output volume.
All settings set to 100% volume. Works fine. No distortion. 
On another workstation which I usually run the volume at 20%, the tones are loud but
not noticeably distorted.
The higher (right) ranges of volume slider seems to not have much gain.  
Anyhow this and the tone are marvelous tools. Thanks.

k1ky:
> I just tried it out on a desktop with typically low output volume.
> All settings set to 100% volume. Works fine. No distortion.
> On another workstation which I usually run the volume at 20%, the tones are loud but
> not noticeably distorted.
> The higher (right) ranges of volume slider seems to not have much gain.
> Anyhow this and the tone are marvelous tools. Thanks.

Can you contact me direct?

gelmce:
The highest frequencies sound harsh. Square wave?
i.e. The higher the frequency the higher the volume.
In the volume slider the settings at the left make a large change in volume and
toward the right make very little change in volume.

dman776:
You can try a different web browser as the sound function is provided by the browser and operating system.

Thanks,
Darryl

k1ky:
Firefox v65.0 does give "some" range (up to around 40% - starts clipping at 50% with no more volume increases above that level ), no good range in Chrome past the 10% level - goes into square wave clipping - clearly overdriven somehow..  Have also experienced total computer lockups in Firefox on 2 different computers (Toshiba Laptop WIN10 v1809 cu.292 twice so far) Nightly-683 Nanostation M2

PR #361, opened 2019-02-10
Adding support for the TP-Link WBS210 version 1.

patrickdickey52761:
My email associated with my callsign is KE0RSX@arrl.net. Changing the $mfgprefix value in admin on line 158 resolves the issue of the filename warning when upgrading the firmware. I'm totally open to other methods of accomplishing this.


ae6xe (approved):
Looks good.

PR #367, opened 2019-02-24
aredn: enable ath9k auto distance capability

ae6xe:


ae6xe:
Tested on several links by 3 different individuals.   hardware tested includes LHG5nD, hAP ac lite, Rocket M3, NSM3-XM, and more.

dman776 (approved):
previously tested ok

PR #372, opened 2019-02-26
feature: standardize browser page title on the node status page

### LINKED ISSUE #366: Node Info TAB name consistency between GUI screens (Remove AREDNtm Node) from Main Node Status Tab
Browser Tab tag names are not consistent when switching between screens.  Preferably would like to see the first few letters (Which are usually Callsign) visible on the TAB and full field description when hovering over the tab. .  The Main Node Status Screen is tagged "AREDN tm Node CALLSIGN......fieldinfo when displaying or hovering over the tab. The MESH STATUS screen shows Callsign....fieldinfo and the  Wi-Fi scan screen tab shows the Field Description (no AREDNtm Node) Same goes for the Setup Screen tab. 

Looks like only the Main Node Status browser Tab needs to have the AREDNtm Node name removed.

dman776:
to be consistent with other pages
fixes #366

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -96,7 +96,7 @@ if($parms{css} and -f "/www/$parms{css}" and $parms{css} =~ /\.css$/i) {
 
 # generate the page
 http_header();
-html_header("AREDN&trade; Node $node", 1);
+html_header("$node status", 1);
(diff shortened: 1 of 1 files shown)


K6AH (approved):
Looks good.

PR #375, opened 2019-02-28
bugfix: delay start up of snmpd until later in the boot process.

### LINKED ISSUE #374: snmpd not starting up properly on some devices
After installing the libnetsnmp and snmpd packages, on certain devices the snmpd service fails to properly start up.  
It starts and is listening, but it does not answer requests.  
Running `/etc/init.d/snmpd restart` will fix it.  

I've been testing/playing around with this on a few of the nodes on our network.  
Two that have this issue are an old Rocket M2 (XM) and a NanoBridge M5.  
I've also heard local chatter of snmpd not starting on a Bullet and an NSM5 (XW).  
I've also tested this on my 5GHz node which is a NanoBeam M5 19, it *does not* have this issue.  

I have, I think, tracked this down to the snmpd service starting up too soon in the boot cycle.
If I change the `START` var at the top of the snmp init.d script to something like "98" *and* also change the `/etc/rc.d/S50snmpd` symlink to be `S98snmpd` then snmpd will start up correctly.

I'm going to just assign this to myself, I think I see how to change it. :)
*edit* Ok, I can't assign it to myself...

r1de:
On some devices snmpd would start before it's configuration file was created.

Changed /etc/rc.d/S50snmpd to /etc/rc.d/S98snmpd

Fixes #374

### DIFF
--- a/patches/713-delay-snmpd-startup.patch
+++ b/patches/713-delay-snmpd-startup.patch
@@ -1,0 +1,12 @@
+Index: openwrt/feeds/packages/net/net-snmp/files/snmpd.init
+===================================================================
+--- openwrt.orig/feeds/packages/net/net-snmp/files/snmpd.init
++++ openwrt/feeds/packages/net/net-snmp/files/snmpd.init
+@@ -1,6 +1,6 @@
+ #!/bin/sh /etc/rc.common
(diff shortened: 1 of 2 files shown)


ae6xe (approved):
looks good

Issue #264, opened 2018-11-10
Auto wifi scan using javascript


raysuelzer:
Hey, 

I'm a front end web developer.  How would you feel about me taking a stab at implementing a javascript based autoscan feature that would update the page via ajax instead of meta-refreshes?  It would probably run faster and require less data over the air.

KM6WUH (soon to be KK6RAY)
Ray

ae6xe:
Ray,  YES.   please contact separately  ae6xe at arrl dot net.   We would love to rebuild a lot of the UI as the current perl generated pages are not very efficient, a memory hog, and don't get us started...  :) .  Let's further discuss.

dman776:
Reviewing this for a new UI approach with Ray.

dman776:
closing.   addressed with #379 and in aredn_ui PR#12

PR #379, opened 2019-03-18
feature: added a wifi scan endpoint to the /api

mcleanra:
this is for #264 , i'll make another PR to aredn_ui shortly with a scan interface that depends on this.  i've only done testing on a nanostation m2


mcleanra:
sent the other related pr: https://github.com/aredn/aredn_ui/pull/12

dman776:
First, thank you for the PR!   Happy to see it.   
what's the thought process behind modifying wscan to add json output, instead of adding code to /api to output scan info (ie.   http://localnode.local.mesh:8080/cgi-bin/api?status=scan      or similar).   The iwinfolua library is available and should contain all the data needed.   It seems that the only reason to add the json scan output is to feed the AngularUI and the /api is the primary data source for that.

mcleanra:
hi @dman776 
that certainly makes sense, thanks for the heads up on the /api approach.  i'm not really that familiar with lua but i took a crack at it and updated the PR by adding a new /api endpoint for the scan page (http://localnode.local.mesh:8080/cgi-bin/api?scan=scanlist).  

i'll modify the angular PR pending your thoughts on this approach

ae6xe:
Ryan,  can you send an email to ae6xe at arrl dot net?   We can setup additional realtime communications to have more interactive dialog on UI changes beyond what we'd want to capture in a github PR.

dman776:
We need to decide how to handle the non-standard channels (ie.   ch254 is ch-2, ch255 is ch-1, 3Ghz channels, etc)....   Need to discuss if we patch iwinfo, libiwinfo-lua, or translate in /api.   Holding pending further discussion.

mcleanra:
We could always expose ```freqlist``` through the API and use that info to do a transform client side with a custom pipe.  Basically the same thing wscan currently does in ```freq_to_chan```

As a bonus we would be able to show what frequencies the device supports, maybe somewhere on the status page

dman776 (approved):
will need to ensure that we create a pipe in the UI to translate non-standard channels to negative channels

PR #386, opened 2019-03-23
feature: added chart endpoint to the /api

mcleanra:
Moved the signal.json code into api so that the new charts page can be built


dman776 (changes requested):
since we are moving this to the api, i think it makes more sense to have two different targets for the charts.   ie.   api?charts=realtime and api?charts=archive

dman776 (approved):
nice work!

PR #387, opened 2019-03-24
enhancement: add additional RF link info to the sysinfo.json file.

r1de:
Added: current TxMbps value (the negotiated modulation rate)
Added: current signal and noise levels
Added: current set frequency

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -176,7 +176,7 @@ end
 -- Return Frequency
 -------------------------------------
 function model.getFreq()
-	local wlanInf=get_ifname(wifi)
+	local wlanInf=get_ifname('wifi')
 	local freq=""
 	freq=os.capture("iwinfo " .. wlanInf .. " info | egrep 'Mode:'")
 	freq=freq:gsub("^%s*(.-)%s*$", "%1")
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -251,13 +251,6 @@ function neighborLinkInfo()
 		neighborLinkInfo[name]["lq"]=lq
 		neighborLinkInfo[name]["nlq"]=nlq
 		neighborLinkInfo[name]["link_type"]=linkType
---		print("Node: "..name)
---		print("TX br: "..tx_rate)
---		print("RX br: "..rx_rate)
---		print("Sig: "..sig)
---		print("Noise: "..nse)
---		print("LQ: "..lq.."% NLQ: "..nlq.."%")
---		print("Link Type: "..linkType)
 	end
 	return neighborLinkInfo
 end
@@ -289,7 +282,6 @@ function local_services()
 		table.insert(lclsrvs, service)
 	end
 	return lclsrvs
---	info['services_local']=lclsrvs
 end
 
 -------------------------------------
@@ -320,7 +312,6 @@ function all_services()
 		table.insert(services,service)
 	end
 	return services
---	info['services']=services
 end
 
 -------------------------------------
@@ -362,12 +353,12 @@ function all_hosts()
 	end
 	info['hosts']=hosts
 end
+
 -------------------------------------
 -- Return link_info (for sysinfo.json)
 -------------------------------------
 function link_info()
 	local linkinfo={}
---	local linkinfo['link_info']={}
 	for name, info in pairs(neighborLinkInfo()) do
 		linkinfo[name]={}
 		for key, value in pairs(info) do
@@ -375,8 +366,8 @@ function link_info()
 		end
 	end
 	return linkinfo
---	info['link_info']=linkinfo
 end
+
 -------------------------------------
 -- Return Channel for Radio
 -- @param radio Radio Device.
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -37,10 +37,12 @@
 require("uci")
 local aredn_uci = require("aredn.uci")
 require("aredn.utils")
+local olsr=require("aredn.olsr")
 -- require("aredn.http")
 local lip=require("luci.ip")
 require("nixio")
 require("ubus")
+require("iwinfo")
 
 -------------------------------------
 -- Public API is attached to table
@@ -173,7 +175,8 @@ end
 -------------------------------------
 -- Return Frequency
 -------------------------------------
-function model.getFreq(wlanInf)
+function model.getFreq()
+	local wlanInf=get_ifname(wifi)
 	local freq=""
 	freq=os.capture("iwinfo " .. wlanInf .. " info | egrep 'Mode:'")
 	freq=freq:gsub("^%s*(.-)%s*$", "%1")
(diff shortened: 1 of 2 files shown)


ae6xe (changes requested):
The bit rate returned is always for a 20MHz channel.  This value is not compensated for in iwinfo with different channel widths.    Need to divide by 2 if 10MHz and by 4 if 5MHz.    

However..  this value doesn't tell you the rate in use on a given link, it is probably only the highest rate in use somewhere.  If 10 links, there could be 10 different rates in use -- this is the transmit rate not receive.   If the purpose is to show rates on links, then you'd need to pull this from:

/sys/kernel/debug/ieee80211/phy0/netdev\:wlan0/stations/BSS addr of remote node/rc_stats

r1de:
Ok. Thanks Joe. I did not know that it was showing some "maximum" bit rate (and I didn't even think of the bit rate for each link, duh!).  
How about we just add the Signal, Noise and Frequency and leave the TxMbps for later?  
There is talk of adding the LQ/NLQ info as well to sysinfo.json, that is also going to be a "per link" item, so it makes sense to have the bit rate there as well in the end.

ae6xe:
A better option would be to pass back the actual throughput of data as show in this table (for each neighbor).   The 'A' row is the actual xmit thoughput (dividing by 2 or 4 for 10 and 5 MHz channels):

/sys/kernel/debug/ieee80211/phy0/netdev:wlan0/stations/MAC address of remote node/rc_stats

All the other measures (SNR, LQ, NLQ) are indirect and 'interesting' and also per neighbor.   If you translated the MAC to an IP address and sent back a table of "to_IP: throughput in mbps", and show on a map link would be the most meaningful.  This value makes different links directly comparable as to their performance.  A 'Quality' link is just one with a value above some throughput threshold.  The only other significant measure is latency of the link.   There's no place to pull this metric from that I know of -- has to be determined with a ping or similar method.

r1de:
We can wait on all of it then.  
These additions to the sysinfo.json were requested by other users, they want to use them in some other scripts or something, I am not certain.  
I was also just trying to get the same values that the GUI had, just for consistency.  

Other than to just display this info and maybe use the frequency to check the band without searching through several channel lists, I really had no idea how I would use this on my map.  

But... you just gave me some other ideas, I think.  
I have to fully grok what you just said. :)  
Those rc_stats tables are quite interesting!  
I know I have seen that info before, but you just made it make way more sense! thanks!

dman776:
Check the PR that adds iwinfo into the new /api page.   That may be a better place for this info.

Thanks,
Darryl

r1de:
@ae6xe @dman776 Please test this, it's worked fine on the few nodes I have tried, but still.

r1de:
Also Dependent on #389

r1de:
I also moved the some of the sysinfo.json info into functions in the aredn/info.lua file.  
`services` and `hosts`  
This will hopefully help later.  

I can't seem to update that latest commit comment, please update if you can.

r1de:
I see a couple of things I might have forgotten. Will update in just a bit.

files/usr/lib/lua/aredn/info.lua
```diff
@@ -171,10 +173,199 @@ function model.getBand(radio)
...
+	freq=os.capture("iwinfo " .. wlanInf .. " info | egrep 'Mode:'")
+	freq=freq:gsub("^%s*(.-)%s*$", "%1")
+	freq=string.match(freq, "%((.-)%)")
+	return freq
+end
+
+-------------------------------------
+-- Return list of MAC to Hostname files
+-------------------------------------
+function mac2host(dir)
```
dman776, line 190: this would make more sense in the aredn.utils library

files/usr/lib/lua/aredn/info.lua
```diff
@@ -171,10 +173,199 @@ function model.getBand(radio)
...
+		list[i] = filename
+	end
+	pfile:close()
+	return list
+end
+
+-------------------------------------
+-- Return Neighbor Link Info
+-------------------------------------
+function neighborLinkInfo()
```
dman776, line 205: this needs to be in the model scope.
ie.   function model.neighborLinkInfo()

files/usr/lib/lua/aredn/info.lua
```diff
@@ -171,10 +173,199 @@ function model.getBand(radio)
...
+		neighborLinkInfo[name]["nlq"]=nlq
+		neighborLinkInfo[name]["link_type"]=linkType
+	end
+	return neighborLinkInfo
+end
+
+-------------------------------------
+-- Return locally hosted services (for sysinfo.json)
+-------------------------------------
+function local_services()
```
dman776, line 261: this needs to be in the model scope.

files/usr/lib/lua/aredn/info.lua
```diff
@@ -171,10 +173,199 @@ function model.getBand(radio)
...
+		service['error']="Cannot read local services file"
+		table.insert(lclsrvs, service)
+	end
+	return lclsrvs
+end
+
+-------------------------------------
+-- Return *All* Network Services
+-------------------------------------
+function all_services()
```
dman776, line 290: this needs to be in the model scope.

files/usr/lib/lua/aredn/info.lua
```diff
@@ -171,10 +173,199 @@ function model.getBand(radio)
...
+		service['error']="Cannot read services file"
+		table.insert(services,service)
+	end
+	return services
+end
+
+-------------------------------------
+-- Return *All* Hosts
+-------------------------------------
+function all_hosts()
```
dman776, line 320: this needs to be in the model scope.

files/usr/lib/lua/aredn/info.lua
```diff
@@ -171,10 +173,199 @@ function model.getBand(radio)
...
+		host['error']="Cannot read hosts file"
+		table.insert(hosts,host)
+	end
+	info['hosts']=hosts
+end
+
+-------------------------------------
+-- Return link_info (for sysinfo.json)
+-------------------------------------
+function link_info()
```
dman776, line 360: this needs to be in the model scope.

files/www/cgi-bin/sysinfo.json
```diff
@@ -147,95 +148,28 @@ info['interfaces']=tif
...
-				service['name']=name
-				table.insert(services,service)
-			end
-		end
-	else
-		service['error']="Cannot read services file"
-		table.insert(services,service)
-	end
-	info['services']=services
+	info['services']=all_services()
```
dman776, line 156: aredn_info.all_services()

files/www/cgi-bin/sysinfo.json
```diff
@@ -147,95 +148,28 @@ info['interfaces']=tif
...
-				service['link']=link
-				table.insert(lclsrvs, service)
-			end
-		end
-	else
-		service['error']="Cannot read local services file"
-		table.insert(lclsrvs, service)
-	end
-	info['services_local']=lclsrvs
+	info['services_local']=local_services()
```
dman776, line 161: aredn_info.local_services()

files/www/cgi-bin/sysinfo.json
```diff
@@ -147,95 +148,28 @@ info['interfaces']=tif
...
+end
+
+-- LOCAL NEIGHBORS LINK INFO
+if string.find(nixio.getenv("QUERY_STRING"):lower(),"link_info=1") then
+	info['link_info']=link_info()
+end
+
+-- MESHMAP INFO
+if string.find(nixio.getenv("QUERY_STRING"):lower(),"meshmap=1") then
+	info['services_local']=local_services()
```
dman776, line 171: aredn_info.local_services()

files/www/cgi-bin/sysinfo.json
```diff
@@ -147,95 +148,28 @@ info['interfaces']=tif
...
+	info['link_info']=link_info()
```
dman776, line 172: aredn_info.link_info()

dman776 (changes requested):
a few inline comments

r1de:
Ok, I see all those and understand.  
Spent all morning trying to get a good firmware build with the lausocket stuff built in... git is a pain sometimes!  

I will resolve them once I get home.

dman776:
Will resubmit with updated PR soon.

PR #389, opened 2019-03-26
feature: add OLSR current neighbor info to API

dman776:
added aredn.olsr lua library to contain OLSR functions
moved some functions from api to aredn.olsr
added aredn.olsr.getCurrentNeighbors to aredn.olsr
added luasocket library to use as http client to fetch json urls


ae6xe (approved):
looks good, via code inspection.

PR #391, opened 2019-03-26
bugfix: add luasocket feed to make luasocket package available

dman776:
added feeds install luasocket to Makefile


ae6xe (approved):
looks good, needs package install to work.

PR #394, opened 2019-03-28
feature: api - added current dhcp leases

dman776:
/cgi-bin/api?services=dhcpleases
returns list of current leases (ip, host, mac)


dman776:
actually, this may be an information leakage issue in the API file.   It may need to move to the "secured" API.   thoughts?

ae6xe:
Yes, this is only under setup today.  As a general practice,shouldn't the rule be to make secure unless reason to open up?

ae6xe (approved):
looks good.  I've not tested runtime.

PR #396, opened 2019-04-03
aredn: Set default channel and width

ae6xe:
10MHz channel width is most commonly used. This default
will minimize how many new users are unable to connect
to a local mesh on first try.  In 2GHz use ch -2
as only clear part 97 10MHz channel available and
predominently used.


dman776 (approved):
good change.  this will be a better set of defaults.

PR #401, opened 2019-04-13
bugfix: ubnt lbe-m5 does not identify boardid

### LINKED ISSUE #398: Ubiquiti Litebeam M5 table of defaults not being used?
Max power, Model name not being used for power selection and status page.  The status page lists:

model: Unknown Hardware 

!!!! UNSUPPORTED DEVICE !!!!
boardid: Ubiquiti Litebeam M5

and I can select a max power higher than 19, even though the perlfunc file says supported=1, and the maxpower option is defined to 19.

Maybe I don't understand how the max power selection works--I am new to this project.

When I change maxpower to something higher, save, and reboot, it seems the device 'reboots' at least twice, and the power ends up back at 19dbm.

Are the table values not applying somehow?

ae6xe:
fixes #398

### DIFF
--- a/files/usr/local/bin/get_boardid
+++ b/files/usr/local/bin/get_boardid
@@ -8,7 +8,7 @@ if [ ! -f /tmp/sysinfo/boardid ]; then
   ar71xx_board_detect
 
   case $AR71XX_BOARD_NAME in
-    *-m-xw)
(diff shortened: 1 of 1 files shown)


dman776 (approved):
easy fix.  looks good.

PR #403, opened 2019-04-15
Fixed the consistency between "archive" and "realtime" SNR data

mcleanra:
I ran into some problems while working on the chart page in [this pull request](https://github.com/aredn/aredn_ui/pull/14).  Our new charts endpoint was causing problems by returning slightly different data when using "realtime" vs. "archive".  

For example, this was the response for a **realtime** SNR data point:

```json 
[
   [
      {
         "tx_mcs":"N\/A",
         "rx_mcs":"N\/A",
         "label":"02:50:31",
         "m":74,
         "y":[
            -21,
            -95
         ],
         "rx_rate":"N\/A",
         "tx_rate":"N\/A"
      }
   ]
]
```

and this was the response for an **archive** SNR data point:

```json
[
   [
      {
         "y":[
            "null",
            -95
         ],
         "x":1555286280000,
         "rx_mcs":"-1",
         "label":"04\/15\/2019 01:58:00",
         "m":"N\/A",
         "tx_mcs":"-1",
         "rx_rate":"0",
         "tx_rate":"1"
      }
   ]
]
```

As you can see, the problems are:

1. the property names are hard to decipher at a glance
2. the returned data is nested inside 2 arrays
3. the string "null" is problematic inside the property ```y```, because sometimes that can be a number (it represents signal in dbm)
4. the "N/A" strings for ```tx_rate```, ```rx_rate```, ```tx_mcs```, and ```rx_mcs``` are problematic, because sometimes they can be numbers
5. the ```label```, which is actually a timestamp, is formatted inconsistently - one includes the date and the other does not
6. the ```m``` property can be either a number or a string
7. the ```x``` property is ```undefined```

With this change we'll get a new response for SNR data that looks like the following - and it will be the same whether it comes from the "archive" (snrlog file) or "realtime" from iwinfo:

```json
[
         {
           "signal_dbm": -28,
           "tx_rate_mbps": 1,
           "timestamp": "2019-04-15T03:59:57Z",
           "noise_dbm": -95,
           "rx_rate_mbps": 0,
           "tx_rate_mcs_index": 0,
           "rx_rate_mcs_index": 0
         }
]
```

Note that ```tx_rate_mcs_index``` and ```rx_rate_mcs_index``` can be undefined.  However, in my opinion that's better than -1 because we can make it obvious in the interface client side.  I renamed each property to make the units more understandable and also got rid of the nested array.  Margin will be calculated client side, and the time stamp will always be in ISO 8601 format in the time zone of the server.

The old signal page is updated to use this new response

### DIFF
--- a/files/www/cgi-bin/api
+++ b/files/www/cgi-bin/api
@@ -98,13 +98,13 @@ function getFreeMemory()
 	return info
 end
 
-function getSignal(realtime) 
+-- gets signal and noise data from either iwinfo (realtime = true) or the files archived in /tmp/snrlog (realtime = false)
+-- the files in /tmp/snrlog are written once per minute by the script /usr/local/bin/snrlog
+function getSignal(realtime)
 	local defnoise=-95
-	local ol={}
 	local dpa={}  -- datapoint table/array
 	local dirname="/tmp/snrlog"
-	local values={}
-	local counter=0
+
 	local datepattern="(%d+)/(%d+)/(%d+)%s(%d+):(%d+):(%d+)"
 
 	local parms={}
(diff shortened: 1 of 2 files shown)


dman776 (approved):
great change to standardize this.

PR #405, opened 2019-04-17
New sysinfo json changes

r1de:
Add additional info about the RF link(s) status to sysinfo.json

Dependent on #389 and #391
Replaces PR #387

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -185,48 +185,6 @@ function model.getFreq()
 end
 
 -------------------------------------
--- Return Neighbor Link Info
--------------------------------------
-function model.neighborLinkInfo()
-	local neighborLinkInfo={}
-	local wlan=get_ifname('wifi')
-	
-	local RFneighbors=iwinfo['nl80211'].assoclist(wlan)
-	local mac2node=mac2host()
-	local hosts_olsr=olsr.getCurrentNeighbors()
-	local name=""
-	
-	for ip, info in pairs(hosts_olsr) do
-		neighborLinkInfo[info["hostname"]]={}
-		for key, value in pairs(info) do
-			-- dont need hostname, we already have it
-			if key ~= "hostname" then
-				neighborLinkInfo[info["hostname"]][key]=value
-			end
-			if key == "linkType" and value == "RF" then
-				for i, mac_host in pairs(mac2node) do
-					local mac=string.match(mac_host, "^(.-)\-")
-					mac=mac:upper()
-					local node=string.match(mac_host, "\-(.*)")
-					if info["hostname"] == node then
-						for stn in pairs(RFneighbors) do
-							stnInfo=iwinfo['nl80211'].assoclist(wlan)[mac]
-							if stnInfo ~= nil then
-								neighborLinkInfo[info["hostname"]]["signal"]=tonumber(stnInfo.signal)
-								neighborLinkInfo[info["hostname"]]["noise"]=tonumber(stnInfo.noise)
-								neighborLinkInfo[info["hostname"]]["tx_rate"]=adjust_rate(stnInfo.tx_rate/1000,bandwidth)
-								neighborLinkInfo[info["hostname"]]["rx_rate"]=adjust_rate(stnInfo.rx_rate/1000,bandwidth)
-							end
-						end
-					end
-				end
-			end
-		end
-	end
-	return neighborLinkInfo
-end
-
--------------------------------------
 -- Return locally hosted services (for sysinfo.json)
 -------------------------------------
 function model.local_services()
@@ -323,20 +281,6 @@ function model.all_hosts()
 		table.insert(hosts,host)
 	end
 	return hosts
-end
-
--------------------------------------
--- Return link_info (for sysinfo.json)
--------------------------------------
-function model.link_info()
-	local linkinfo={}
-	for name, info in pairs(model.neighborLinkInfo()) do
-		linkinfo[name]={}
-		for key, value in pairs(info) do
-			linkinfo[name][key]=value
-		end
-	end
-	return linkinfo
 end
 
 -------------------------------------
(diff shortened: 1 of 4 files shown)


r1de:
Note:
Check the display of links that are dtd, tun, etc.
Seems to only show RF links now. (which is just fine)
I think the intended behavior was different though.

dman776:
the olsr.getCurrentNeighbors() function is returning linkType properly....

`
> aredn_olsr=require("aredn.olsr")
> ho=aredn_olsr.getCurrentNeighbors()
> print(ho)
table: 0x83be80
> require("aredn.utils")
> print_r(ho)
table: 0x83be80 {
  [10.114.181.21] => table: 0x83be80 {
                       [neighborLinkQuality] => 1
                       [linkQuality] => 1
                       [hostname] => "K5DLQ-UNSM2-101.local.mesh"
                       [olsrInterface] => "wlan1"
                       [linkType] => "RF"
                     }
  [10.115.181.21] => table: 0x83be80 {
                       [neighborLinkQuality] => 1
                       [linkQuality] => 1
                       [hostname] => "K5DLQ-UNSM2-101.local.mesh"
                       [olsrInterface] => "eth1.2"
                       [linkType] => "DTD"
                     }
}
`

r1de:
Ok, I am using the function wrong then or something. I'll try to look tonight.

Thx

r1de:
I see where *my* confusion was coming from now.
The `neighborLinkInfo()` function in info.lua is *only* looking for the RF neighbors.  
I will have to change it to show *all* of the neighbors and add in the additional info for the RF neighbors.  

So, @dman776 and @ae6xe what if we were to use this neighbor link info in sysinfo.json file to **replace** the need for the "dot_draw" OLSR plug-in?  
Since the same info (and more) would now be available via the sysinfo.json file, the dot_draw plugin would be redundant.    
That would rid us of the plug-in itself *and* the need for xinetd, since, IIRC, xinetd is only there to redirect port 2004 to localhost:2003.  

My map uses port 2004 to get quite a bit of info, I can change that, but I am not sure if any other people are using the info on port 2004 for anything, perhaps we could make a sticky on the forums asking if anyone is using the port and go from there?  

Just thinking out loud here.  
I'll fix this PR so it works "correctly" regardless...

r1de:
misclick... sorry. :)

r1de:
Tested on 3 different nodes, 1 with many RF links, 1 with many DtD links, and 1 that is only PtP.  
Not tested with tunnels yet.

dman776:
isn't this information already in the olsr.getCurrentNeighbors() function?  Why not just call it instead of creating another function in info.lua?

r1de:
Not *all* of the info is in there, but most of it. The additional RF link info gets added in by my function.  
But you know, I was asking myself this same thing earlier, and I am not sure. :smile:   
I think part of it was having to add the additional info in for the "RF" linkType and another part is just a personal preference of having the nodes name as the "index" instead of the IP.  

I'll look it over again, I probably can be more efficient.

dman776:
let's combine if we can.  Since it's json it shouldnt break anything by adding new data to the structure.

dman776:
also, if your code adds additional processing load (ie. file i/o, etc.), you can add an argument to the function to enable/disable your additional data.  just a thought

r1de:
Good idea(s).  
I'll look once I get out of puppy chasing mode...   "Nooooo, don't chew on that..." :smile:

r1de:
This *is* better I think.

dman776 (changes requested):
when i copy these changed files to a node, sysinfo.json fails with:
root@K5DLQ-MRB-HAPACLITE:/www/cgi-bin# ./sysinfo.json
/usr/bin/lua: /usr/lib/lua/aredn/info.lua:69: attempt to index upvalue 'aredn_uci' (a boolean value)
stack traceback:
	/usr/lib/lua/aredn/info.lua:69: in function 'getNodeName'
	./sysinfo.json:62: in main chunk
	[C]: ?
root@K5DLQ-MRB-HAPACLITE:/www/cgi-bin#

r1de:
Recopied scripts to a node I was using for testing...
If I run the new scripts from the command line like that, I get:
```
root@WD6EBY-VC-SulphurMtn-to-Oxnard-5G:/www/cgi-bin# ./sysinfo.json 
/usr/bin/lua: ./sysinfo.json:151: attempt to index a nil value
stack traceback:
	./sysinfo.json:151: in main chunk
	[C]: ?
```
The "QUERY_STRING" is nil
Running from httpd works fine though.

I'm not even sure how you got that error.

r1de:
@dman776 I still cannot recreate your error.
I modified my script to continue if the "QUERY_STRING" is nil, the script runs fine now and outputs as it should, but I still do not get that error when I run sysinfo.json from the command line.
Are you sure the firmware on your K5DLQ-MRB-HAPACLITE system has some of the newer updates needed? (#389 and #391)

dman776:
just to be sure, do a build and load it to test.

r1de:
roger roger

dman776:
to ensure a clean build....

docker run --rm -it --name builder arednmesh/builder

then, inside the container:

git pull
git fetch origin pull/405/head:pr-405
git checkout pr-405
make

r1de:
I still can't recreate it.
```
root@kg6wxc-test:/www/cgi-bin# cat /etc/mesh-release 
wxc-pr-405-e9c06dc
```
```
root@kg6wxc-test:/www/cgi-bin# ./sysinfo.json
/usr/bin/lua: ./sysinfo.json:151: attempt to index a nil value
stack traceback:
	./sysinfo.json:151: in main chunk
	[C]: ?
```

If I change the script on the node so it continues even if QUERY_STRING is nil, it works just fine and there are no other errors.
```
root@kg6wxc-test:/www/cgi-bin# ./sysinfo.json
Content-type: application/json
Cache-Control: no-store
Access-Control-Allow-Origin: *


{
   "lon": "-120.360363",
   "sysinfo": {
     "uptime": "0 days, 0:04:49",
     "loads": [
       0.46999999999999997,
       0.40000000000000002,
       0.19
     ]
   }, ...
```

dman776:
Starting a container build on my side now just to be sure.

dman776 (approved):
tested ok with a local build.

r1de:
Should I also add in the changes to make it run from the command line when there is no "QUERY_STRING"?  
Do we need it to run from the command line?  
I guess it could be handy for something... :)

dman776:
Sure

Thanks,
Darryl

PR #407, opened 2019-04-23
Add experimental support for the GL.iNet GL-AR150

ZL2WRW:
I was surprised at how easy it was to build AREDN firmware for the GL.iNet GL-AR150
These devices are available from the usual sources for $20 - $30 USD or so - an economy mesh node.
https://openwrt.org/toh/gl.inet/gl-ar150

Mesh networking and tunnelling have been confirmed working.
This device has also been confirmed capable of meshing on channel -1, but /etc/config/wireless has to be manually edited to select this channel because the AREDN web UI does not allow it to be selected (unlike the UBNT Rocket M2 etc).

Following is from: http://zl2wrw-65-83-239.local.mesh:8080/cgi-bin/sysinfo

 node: ZL2WRW-65-83-239
model: Unknown Hardware 

!!!! UNSUPPORTED DEVICE !!!!
boardid: GL.iNet GL-AR150
Device has not been tested. Please file a ticket with your experiences.

br-lan E4:95:6E:42:53:EF
eth0   E4:95:6E:42:53:EF
eth0.1 E4:95:6E:42:53:EF
eth0.2 E4:95:6E:42:53:EF
eth1   E4:95:6E:41:53:EF
tunl0  00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
wlan0  E4:95:6E:41:53:EF
wlan0-1 E6-95-6E-41-53-EF-00-00-00-00-00-00-00-00-00-00

/proc/cpuinfo
system type		: Atheros AR9330 rev 1
machine			: GL.iNet GL-AR150
processor		: 0
cpu model		: MIPS 24Kc V7.4
BogoMIPS		: 265.42
wait instruction	: yes
microsecond timers	: yes
tlb_entries		: 16
extra interrupt vector	: yes
hardware watchpoint	: yes, count: 4, address/irw mask: [0x0ffc, 0x0ffc, 0x0ffb, 0x0ffb]
isa			: mips1 mips2 mips32r1 mips32r2
ASEs implemented	: mips16
shadow register sets	: 1
kscratch registers	: 0
package			: 0
core			: 0
VCED exceptions		: not available
VCEI exceptions		: not available

nvram
hsmmmesh.settings=settings
hsmmmesh.settings.wifimac='e4:95:6e:41:53:ef'
hsmmmesh.settings.mac2='65.83.239'
hsmmmesh.settings.dtdmac='65.83.239'
hsmmmesh.settings.config='mesh'
hsmmmesh.settings.node='ZL2WRW-65-83-239'

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ The Mikrotik hAP AC Lite, Ubiquiti AirRouter, and AirRouter HP are pre-configure
 
 The GL.iNet GL-AR150 is pre-configured with the following VLANS:
 
-* Port labeled "WAN": untagged = AREDN LAN, vlan 1 = AREDN WAN, vlan 2 = DtDLink (device to device)
+* Port labeled "WAN": untagged = AREDN WAN
 * Port labeled "LAN": untagged = AREDN LAN, vlan 2 = DtDLink (device to device)
 
 ## Submitting Bug Reports
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Mikrotik RBLHG-5HPnD | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik RBLHG-5HPnD-XL | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik LDF-5nD | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik QRT5 RB911G-5HPnD-QRT | mikrotik-nand-large | 64Mb | stable
-GL.iNet GL-AR150 | gl-ar150 | 64Mb | EXPERIMENTAL
+GL.iNet GL-AR150 | gl-ar150 | 64Mb | stable
 
 Latest Mikrotik installation options are found at: https://www.arednmesh.org/content/installation-instructions-mikrotik-devices
 
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Mikrotik RBLHG-5HPnD | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik RBLHG-5HPnD-XL | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik LDF-5nD | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik QRT5 RB911G-5HPnD-QRT | mikrotik-nand-large | 64Mb | stable
+GL.iNet GL-AR150 | gl-ar150 | 64Mb | EXPERIMENTAL
 
 Latest Mikrotik installation options are found at: https://www.arednmesh.org/content/installation-instructions-mikrotik-devices
 
--- a/README.md
+++ b/README.md
@@ -135,6 +135,11 @@ The Mikrotik hAP AC Lite, Ubiquiti AirRouter, and AirRouter HP are pre-configure
 * Port 5: DtDLink Port Mesh Routing -- Connect to another mesh node or 8021.q switch. Packets in/out of this port must be vlan 2 tagged, other packets are ignored.
 * Ports 2-4: LAN devices -- Packets in/out of this port are expected to be untagged. The mesh node will (default) DHCP assign an IP address to your computer, ipCam, voip phone, etc. connected to these ports.
 
+The GL.iNet GL-AR150 is pre-configured with the following VLANS:
+
+* Port labeled "WAN": untagged = AREDN LAN, vlan 1 = AREDN WAN, vlan 2 = DtDLink (device to device)
+* Port labeled "LAN": untagged = AREDN LAN, vlan 2 = DtDLink (device to device)
+
 ## Submitting Bug Reports
 
 Please submit all issues to http://github.com/aredn/aredn_ar71xx/issues
--- a/files/www/cgi-bin/perlfunc.pm
+++ b/files/www/cgi-bin/perlfunc.pm
@@ -984,7 +984,7 @@ sub hardware_info
   %model = (
     'GL.iNet GL-AR150' => {
       'name'            => 'GL.iNet GL-AR150',
-      'comment'         => 'EXPERIMENTAL',
+      'comment'         => '',
       'supported'       => '1',
       'maxpower'        => '18',
       'pwroffset'       => '0',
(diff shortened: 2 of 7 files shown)


ae6xe:
ZL2WRW,  there are a couple of places to edit and be fully AREDN functional.  vlan definitions (dtdlink probably not defined) and detecting the model # to have the right power and channel selection options in basic settings.   I'll take a look at shortly to give some big pointers what this code should be so we can include these changes now.

dman776:
as well as additions to /www/cgi-bin/perlfunc.pm

ZL2WRW:
Hi Joe,

Thanks, I'd appreciate some big pointers on where to look as I am new to 
AREDN.

73 ZL2WRW
Ross Whenmouth <ross@topwire.co.nz>

README.md
```diff
@@ -97,6 +97,7 @@ Mikrotik RBLHG-5HPnD | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik RBLHG-5HPnD-XL | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik LDF-5nD | mikrotik-rb-nor-flash-16M | 64Mb | stable
 Mikrotik QRT5 RB911G-5HPnD-QRT | mikrotik-nand-large | 64Mb | stable
+GL.iNet GL-AR150 | gl-ar150 | 64Mb | EXPERIMENTAL
```
dman776: remove the EXPERIMENTAL and change to stable

files/www/cgi-bin/perlfunc.pm
```diff
@@ -982,6 +982,15 @@ sub hardware_boardid
 sub hardware_info
 {
   %model = (
+    'GL.iNet GL-AR150' => {
+      'name'            => 'GL.iNet GL-AR150',
+      'comment'         => 'EXPERIMENTAL',
```
dman776: remove EXPERIMENTAL comment

ZL2WRW:
I've made an entry to www/cgi-bin/perlfunc.pm for the gl-ar150 and the UI seems to handle channel selection, etc properly now.

With reference to page 313 of: https://www.eeboard.com/wp-content/uploads/downloads/2013/08/AR9331.pdf
The AR9331 should be capable of 20 dBm TX. However, whilst the UI allows 20 dBm to be selected when 'maxpower' => '20' is set in perlfunc.pm, "iwinfo" and "iw wlan0 info" show that txpower is capped at 18 dBm. So for the time being I have set 'maxpower' => '18'.

LAN/host direct (untagged), and vlans for WAN (tagged 1) and DTDLINK (tagged 2) seem to be working properly via the Ethernet socket (eth0) labelled WAN on the plastic housing of the gl-ar150. I have confirmed the operation of the WAN (tagged 1) by tunnelling to ZL2ARN and of DTDLINK (tagged 2) with a UBNT Rocket M2 (3.19.3.0 stable).

The 2nd Ethernet socket (eth1) labelled LAN isn't configured for anything at this stage. So, for consistency with other dual-Ethernet AREDN devices, what should the default configuration for eth1 be?

files/www/cgi-bin/perlfunc.pm
ZL2WRW: Done.

ae6xe:
Power settings:  Openwrt settings of power account for both the AR93xx power level, but also an external Power Amp.   One option is to look at the manufacture specifications of the max xmit power (at the lowest MCS rate) and use this for the max setting.   There is also going to be vendor flash data that lists regulatory limits and max settings.   On TPlink devices, trying to set the power above these limits doesn't work -- it can be different limits based on the channel.    Try different channel settings and test power settings, use "iw dev wlan0 info" to see what txpower setting the wireless driver accepted.   same max on all channels, different by channel?     On Ubiquiti devices there is a hidden 'offset'.  the iw command will show you a value, then you have to add the offset to determine actual xmit power.  You'd see this if max specs and max iw value don't match.

vlans and port usage:   I'd propose configuring the ports this way...   Port labeled WAN is configured the same (untagged).    Port labeled LAN is both (LAN-untagged and DtDLink-vlan2).  This means traveling and using this device would not require another switch for typical usage in the hotel to connect to internet and tunnel.  right now only the eth0 port is configured for LAN, WAN, DtDLink (I had put in settings in anticipation of future device support :) ).  eth1 is not configured.   I'll send you an updated patches/708-define-aredn-networks.patch file to test out and submit (tomorrow).

We'll need to review the LED status light behavior to see if any changes are needed too.  Can you attach a support data download from Admin page?  I'd like to review the dmesg and other information to make sure no other issues are found.

ae6xe:
wl2wrw,  update the patches/708-define-aredn-networks.patch file with the following to configure vlans:

https://drive.google.com/file/d/1USqETHnxq3NRKKEBZU9OE3_-EmjfFKy6/view?usp=sharing

The README.md file will need updated in the section "Ethernet Port Usage" to define that this device's  WAN port is the AREDN WAN port and the LAN port is both AREDN LAN and DtDLink port.

ae6xe:
This device doesn't have any Signal Strength LEDs to configure -- typically the low signal LED is used to show where there is an RF, DTDLink, or tunnel, neighbor link.  So shouldn't be any work to do here.   Give the ports a test and see if WAN and LAN are working (your laptop can browse the internet).  Ideally, dtdlink with other mesh devices.  If all is working, should be good to go after the README update.

ZL2WRW:
WAN & LAN ports are working as expected, as is DtDLink and tunnelling.

However, LAN & WAN link/activity LEDs are transposed, but this appears to be an upstream issue with OpenWRT. There is a 3rd (orange) LED which I would like to make work as a "mesh connected" LED (like the red LED on the UBNT Rocket M2). I am going to have a go at patching aredn_ar71xx/openwrt/target/linux/ar71xx/base-files/etc/board.d/01_leds

[supportdata-ZL2WRW-QTH-201904250818.zip](https://github.com/aredn/aredn_ar71xx/files/3116237/supportdata-ZL2WRW-QTH-201904250818.zip) tar file inside a ZIP because Github rejects the .tgz that AREDN supportdata link provides

ae6xe:
check out /usr/local/bin/linkled.   Would likely need an edit in this file for the 3rd light (the LED logic isn't integrated well with 01_leds, yet).   In an ideal world, we'd patch 01_leds one place for all devices for all AREDN needs, TBD.   This is same approach as 02_network as dtdlink definition has been added in one place.

ZL2WRW:
My GL-AR150 seems to be working great now. Is there anything more that I need to change before you will be happy to merge my pull request?

ae6xe (approved):
code inspection, looks good.

PR #408, opened 2019-05-02
bugfix: fix call to getFrequency() to getFreq() in api

r1de:
getFrequency() becomes getFreq()

I found this while I was getting the new UI test setup.


dman776 (approved):
next time, provide a more detailed description. (i fixed this one).   Good catch.

PR #411, opened 2019-05-05
aredn: add support for GL-USB150 Microuter

ae6xe:


ae6xe:
Tested on win10.  Plugged into USB, found new network interface.  browsed to IP address of device and in the GL menu, uploaded AREDN firmware.    Once device boots, browse to standard AREDN IP addresses (making sure you aren't connected to the mesh or same IP address on another network port).     No LED's to configure.  Only 1 LED blinks during boot and solid after boot.  It is configured with vlan 1 and 2, so if a device could configure these vlans and pull out the respective networks.

dman776 (approved):
looks good.  very straightforward

PR #412, opened 2019-05-05
bugfix: fix index error in aredn_info lib

dman776:
getNodeName and getNodeDescription should use [1] instead of [0]


wu2s (approved):
Looks good.

PR #413, opened 2019-05-06
/apiprotected post/validation data

dman776:
basic setup page validators


dman776:
channel "band" is still hard coded to 2400 until we get the board.json/aredn.json setup file with the band id in it.

PR #417, opened 2019-05-10
feature: add "MESH" led for GL USB150

dman776:
enabled the green LED towards the "wifi" end of the device to be the "MESH" active indicator.


K6AH (approved):
Looks good

PR #420, opened 2019-05-18
bugfix: add build config for ar300m

ae6xe:


dman776 (approved):
approved.  merging now.

PR #421, opened 2019-05-18
bugfix: Mesh status missing owning host of service advertisement

ae6xe:
Mesh status will show service entries of a remote mesh
node at the top with 0 ETX value.  The owning mesh hostname
displays blank across the mesh network.

The root cause is OLSR not propagating the node's hostname due
to the admin assigning  a conflicting/duplicate hostname.  These
services will no longer show in mesh status.


dman776 (approved):
Looks good

PR #422, opened 2019-05-18
aredn: check less frequently to update Mesh link LED

ae6xe:
OLSR has rare symptoms of live lock and stops
updating routing and host information.  Suspect cause is
single threaded implementation with plugins continually
called from mutiple sources.  Prior fixes to ensure
read/write calls are properly written may not have
fully resolved the causes.  This change will reduce calls
to retrieve olsr data.


dman776 (approved):
approved

PR #432, opened 2019-06-19
bugfix: Fix instance of sysupgrade failing

ae6xe:
It was discovered that sysupgrade would exit with an
error when some processes could not be killed.  The
wifi interface is brought down to avoid this condition
prior to doing the sysupgrade


dman776 (approved):
looks good

PR #433, opened 2019-06-19
bugfix: wan interface fails to function on UBNT NS XM devices

ae6xe:
The WAN network fails to be configured correctly on
UBNT NS XM devices.


dman776 (approved):
good catch

PR #440, opened 2019-07-02
readme: updated with note regarding gl.inet keep settings

### LINKED ISSUE #439: Add note in README for GL.iNet devices
Add a note in README for GL.iNet devices, advising to uncheck the "Keep Settings" box prior to installation of AREDN firmware.

dman776:
fixes #439


ae6xe (approved):
Looks good.

Issue #337, opened 2019-01-26
WiFi Scan


ZL2ARN:
Firmware 3.16.1.0 provided free access to WIFI scans on the Mesh Nodes.
All later versions now require "password" access to see the scan list.
My memory is a little poor on why this was changed. Was it for security reasons. ??
We have an extensive WIFI Network here in Wellington ZL, and it is vital to use the WiFi Scan at each Node to check for adjacent activity, and to prevent any Channel clashes.
Would it be possible to re-instate **no password** access again ??

ae6xe:
The wifi scan will trigger RF emissions on all the channels.   On 5GHz this would include DFS channels where you might have primary doppler radar in your area.  With abundance of caution, the password ensures that only the node owner or trusted users can control the emissions.

ZL2ARN:
Hi Joe,
Thanks for your reply.
“A client can use two scanning methods: **active** and **passive**. During an active scan, the client radio transmits a probe request and listens for a probe response from an AP. With a passive scan, the client radio listens on each channel for beacons sent periodically by an AP.”
Can you confirm it is an **active scan** in the Mesh Nodes ?? 
The speed of reply would suggest it is an active scan. A passive scan would I think take much longer waiting for the periodical beacons to occur.

ZL2ARN:
Thanks Joe,
I have very carefully edited  ../bin/wscan in the two locations as you detailed. Then rebooted just to ensure no scan data was saved in memory.
First impressions are that nothing has changed and WIFI scan results are just the same.
So will now locate a Spectrum analyser to observe (RF) both active and passive scans.
My test node is just a few feet away from the work station so will be able to see any active scans.

ae6xe:
With a specturm analyser, the ms blip of energy wouldn't be noticeable.     The way I'd do this is capture packets on the neighbor node, then review the traffic to see what was transmitted or not.  To do this, install the "tcpdump-mini" package (you may have to load the current nightly build to do so, given signatures on the packages).  

Run this command, "tcpdump -i wlan0-1 -c 1000 -w /tmp/scan.pcap"

[...]

PR #444, opened 2019-08-25
aredn: switch from active to passive wifi scan

ae6xe:
Password protection is no longer necessary with
a passive scan.  results are same or very similiar,
both options do not always show all devices.


ae6xe:
Issue request:  https://github.com/aredn/aredn_ar71xx/issues/337

dman776 (approved):
Looks good

PR #454, opened 2019-09-25
cleanup: remove the `meshmap=1` option from sysinfo.json.

r1de:
It is no longer needed and never actually was.
You learn something new every day. :)

Every bit saved is a bit earned for something else.


dman776:
Did we ever do an official release with this?  I haven't gone back to look yet.   If so, we should increment the version number.

PR #458, opened 2019-10-11
aredn: add definition for Mikrotik SXTsq 2nD

ae6xe:
tested by K9CQB <damon.schaefer@me.com>


dman776 (approved):
straight forward change.  looks good.

PR #459, opened 2019-10-27
Reduce visual clutter on mesh status page

kostecke:
This change reduces visual clutter on the mesh status page by removing redundant _local.mesh_ domains from link text.

### DIFF
--- a/files/www/cgi-bin/mesh
+++ b/files/www/cgi-bin/mesh
@@ -376,8 +376,9 @@ if(keys %localhosts)
     foreach $ip (keys %localhosts)
     {
 	$host = $localhosts{$ip}{name};
+        $localpart = $host =~ s/.local.mesh//r;
 	$tactical = $localhosts{$ip}{tactical} ? " / " . $localhosts{$ip}{tactical} : "";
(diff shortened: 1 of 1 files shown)


r1de:
What about dropping anything after 30 characters or so and changing the text to `...`?

k6ccc:
I have no issue with dropping the .local.mesh from the text (less clutter is good), but DO leave .local.mesh in the link to the node.  Not likely an issue for computers that are only on the AREDN network, but a big one for computers that are also on another network.

r1de:
Yes, in this PR, the links retain the `.local.mesh` (as they *always* should (for the above reasons))  
It's just a cosmetic change, as would be the `...` I was talking about.

ae6xe:
This change could be standalone, adding value by itself.  To truncate the hostnames in mesh status, could be a separate commit, but I'd propose adding an advanced setting to define the length the node admin can configure.

kostecke:
Here's a comparison of the left columns of my patched and the current versions of the Mesh Status page for anyone who has not tried the new version. My call-sign is the only one left readable.

k0stk-mesh-status-patch-comparision

ae6xe (commented):
Looks good.

kostecke:
Thank-you!

PR #460, opened 2019-10-30
bugfix: +x to /www/cgi-bin/mesh - 403 error if not.

r1de:
Just noticed this...


ae6xe:
good catch, testing of file without a full image and install didn't catch this.

ae6xe (approved):
puts back to same file permissions as before.

PR #461, opened 2019-11-03
aredn: add definition for Mikrotik RBSXTsq2nD

ae6xe:
Mikrotik has multiple model strings for this device


dman776 (approved):
Looks good.  Thanks Joe

PR #467, opened 2019-12-02
aredn: common build config updates

ae6xe:
remove ZRAM option causing boot errors and not used
remove unused olsr secure plugin from image to save space
add option to build blockknownencrytion aredn package


dman776 (approved):
Good change

PR #473, opened 2020-01-29
aredn: upgrade to Openwrt 19.07.0 release

ae6xe:
Co-authored-by: Andrew KK4ZUZ <apcameron@softhome.net>

This commit begins the process to migrate from ar71xx to
ath79 target images.


dman776 (approved):
tested ok

PR #474, opened 2020-01-30
aredn: drop SSL from tunnels

ae6xe:
installing tunnels brings along a 900k+ SSL package
yet encryption is turned off. To enable 8M Flash devices
to continue to install tunnels, this dependency is removed.


ae6xe:
superceded with another PR.  This is no longer needed.

PR #475, opened 2020-01-30
feature: remove libopenssl (large library) and disable ssl in vtun

dman776:
save space in flash by not requiring the massive libopenssl library for vtun.


ae6xe (approved):
looks good.  Separately tested this option and package install succeeds without dependency to openssl package.

PR #478, opened 2020-02-10
bugfix: fix the Authentication for vtun to be compatible with the pre…

apcameron:
bugfix: fix the Authentication for vtun to be compatible with the previous tunnels


r1de:
:+1:

dman776 (approved):
tested successfully by Eric.

PR #480, opened 2020-02-14
enhancement: add throughput info for each link to sysinfo.json

r1de:
when `sysinfo.json` is called with `link_info=1` throughput info will be added for each link.
This runs grep against `/proc/net/dev` for each _linked_ interface over a 1 second interval.
it does slow down the response time of sysinfo.json, depending on how many linked interfaces there are.
I have many active tunnels and have not had any issues.

output is a string "RX TX" in **bytes** for that 1 second interval.

This could be useful for local graphing or other things.
I have some ideas to use it for the meshmap in the future.

### DIFF
--- a/files/usr/lib/lua/aredn/info.lua
+++ b/files/usr/lib/lua/aredn/info.lua
@@ -44,6 +44,7 @@ require("nixio")
 require("ubus")
 require("iwinfo")
 
+os.capture=capture
 -------------------------------------
 -- Public API is attached to table
 -------------------------------------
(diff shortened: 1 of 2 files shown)


r1de:
will convert the "RX TX" string into 2 separate json values...

r1de:
Actually, hold off on this... I can see situations where this would take close to 30 sec's to complete...
That is way way too long and will cause timeouts...

dman776:
closing for now.

PR #481, opened 2020-02-20
bugfix: ath79 tplink cpe210 v2 changed name

ae6xe:


dman776 (approved):
verified via forum user

PR #482, opened 2020-02-23
bugfix: added missing network defintions for gl-ar150

ae6xe:


ae6xe:
change tested by KC0ODE successfully.

PR #486, opened 2020-02-24
enhancement: Download paths

r1de:
Add ability to change the firmware/packages download paths and also change the "internet check" so that if the URL contains ".local.mesh" then it will try to download the needed info from the mesh host.

Tested pretty good, but still review please.


dman776 (changes requested):
the firmware download path changes are in a separate PR.   Please remove that from this PR.

r1de:
I don't understand what the conflict is showing or how to resolve.
As I understand, when #488 was merged, it changed the default firmware download location and if my #486 was merged later it would have overwritten that change.
There is a conflict now that I fixed that though?

I do not get git sometimes. :slightly_frowning_face: 

Sorry if I broke something

r1de:
I broke this. closing and recreating.

PR #487, opened 2020-02-26
aredn: upgrade to pre-openwrt-19.07.2

ae6xe:


ae6xe:
team decision to push this commit with top of openwrt 19.07 branch and make available to keep testing close to imminent 19.07.2 release.

PR #488, opened 2020-02-26
feature: change firmware download location

dman776:
change download location from /firmware/ubnt to /firmware


ae6xe (approved):
looks good.

PR #490, opened 2020-03-04
enhancement: ability to change and revert firmware and package downlo…

### LINKED ISSUE #452: Enhancement request: Mesh repository & Maps
To make upgrade process of AREDN simplier and less bandwidth hungry, we have decided to run our own AREDN mirror that is accessible over the mesh. This mirror looks monthly for new version of AREDN (From the stable repo only, not the nightly).

So we changed the paths in the advance configuration section to reflect our mesh server. The problem is that nodes return «Error: no route to the Internet» when trying to refresh the firmware list. The map on the «Basic Setup» doesn't appear. This behavior is observed on nodes without internet access.

If the internet is available, the data (map tiles, leaflet.js, available firmware version...) are fetched from our mesh server as we wished.

It seems obvious to me that there is a «safety» switch in the code to ensure no hanging requests are made. This safety mechanism should detect mesh link and proceed from the mesh even if internet is not available.

Screenshot joined. All devices except the GL-USB150 runs 3.19.3.0.

2019-09-22 13_51_41-VA2XJM-GLUSB150 administration
2019-09-22 13_52_40-VA2XJM-GLUSB150 Advanced Configuration
2019-09-22 14_04_58-VA2XJM-GLUSB150 setup

r1de:
…ad paths

replaces #486
fixes #452

### DIFF
--- a/files/etc/init.d/local
+++ b/files/etc/init.d/local
@@ -74,4 +74,25 @@ boot() {
     usbvalue=$duval
   fi
   /usr/local/bin/usb_passthrough "${usbvalue}"
+
+  # package repositorys
+  local repos="core base arednpackages packages luci routing telephony"
+  set -- $repos
+  while [ -n "$1" ]; do
+    local ucirepo=$(uci -q get aredn.@downloads[0].pkgs_$1)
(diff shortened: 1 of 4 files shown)


dman776 (approved):
tested ok by K5DLQ

PR #497, opened 2020-04-01
bugfix: Correct CPE510 v3 Image

apcameron:
Tested by Bill NG1P
See https://www.arednmesh.org/content/tp-link-cpe510-v38-support


ae6xe (approved):
looks good.

PR #498, opened 2020-04-01
aredn: apply aredn port config settings to NS XW ath79

ae6xe:


dman776 (approved):
tested by AE6XE successfully

PR #500, opened 2020-04-03
feature: add target type info (ar71xx/ath79) to admin page

dman776:
add firmware type to admin page


ae6xe (approved):
looks good.   optional -- suggest to change order, e.g. "(ath79/generic) (nanostation-m)"  so this flows nicely with where these strings are in the pathname to the image file in downloads.

Manually edited on a node and tested.

PR #501, opened 2020-04-04
bug: leaflet css and js files from mesh hosts

r1de:
forgot about the map on the setup page in PR #490
This fixes it.


dman776 (approved):
look good

PR #505, opened 2020-04-07
feature: add "aredn alerts" feature in header

dman776:
Provide basic message support for nodes that have internet access.  The AREDN team can push a message to ALL nodes, or, to individual nodes (by nodename), or both.

### DIFF
--- a/files/etc/crontabs/root
+++ b/files/etc/crontabs/root
@@ -2,4 +2,5 @@
 * * * * * /usr/local/bin/rssi_monitor
 * * * * * /usr/local/bin/snrlog
 * * * * * /usr/local/bin/clean_zombie.sh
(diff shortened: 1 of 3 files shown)


ae6xe (changes requested):
suggest optimizing the code by removing the if then else constructions where possible.  

From:

ping -W10 -c1 downloads.arednmesh.org > /dev/null
if [ $? -eq 0 ]
then
if [ -f /tmp/aredn_message ]
then
rm /tmp/aredn_message
fi

to (better check this logic/syntax, but the idea is to use || and &&):
[ ping -W10 -c1 downloads.arednmesh.org > /dev/null ] || [ \! -f /tmp/aredn_message ] || rm /tmp/aredn_message

ae6xe (approved):
looks good

PR #506, opened 2020-04-07
bugfix: fix variables in perlfunc. fixes #504

### LINKED ISSUE #504: Mikrotik Nightly 1445 Save Changes Issue
Noticed this on a Mikrotik hAP rb-952ui-5ac2nd after changing the dtd.

Scalar value @target[1] better written as $target[1] at /www/cgi-bin/perlfunc.pm line 2059.
Scalar value @release[1] better written as $release[1] at /www/cgi-bin/perlfunc.pm line 2070.
Scalar value @_[0] better written as $_[0] at /www/cgi-bin/perlfunc.pm line 2089.
Scalar value @nums[0] better written as $nums[0] at /www/cgi-bin/perlfunc.pm line 2097.
Scalar value @nums[1] better written as $nums[1] at /www/cgi-bin/perlfunc.pm line 2097.

This also seems to be causing issues with olsrd because when I revert back to 3.20.3.0 I don't have issues with routing again. With Nightly 1445 it can't seem to route via node name, only IP addy.

On a hunch I just hit Save Changes on the Administration page on an LDF5, hAP and Basebox5 and I  get the same Scalar value errors as well.

All of my Mikrotik Nodes are now on 3.20.3.0

Thanks
Dale, KC2CBD

r1de:


dman776 (approved):
looks better!

PR #510, opened 2020-04-09
bugfix: Migrate the TP-LINK CPE510 v2/v3 to the ath79 branch

apcameron:
Switch to the ath79 branch to work around the bug in the ar71xx branch with the Ethernet drivers.
Tested by N3XKD


ae6xe (approved):
looks good.  User tested in forum dialog.

PR #511, opened 2020-04-10
feature: add aam_refresh method to advanced config page

dman776:


ae6xe (approved):
look good.

Orv:
What does this do?

ae6xe:
It is a companion to this feature:  https://github.com/aredn/aredn_ar71xx/pull/505

ae6xe:
In anticipation, #505 is a feature to enable an AREDN mechanism to display notices or announcements globally and/or by individual node name to mesh nodes that have access to the internet.    The node will check if there is access to the internet, if so, then download these messages and display in node status.

Orv:
Thanks.

Orv

dman776:
FYI, for the AREDN Alert Message system:   https://arednmesh.readthedocs.io/en/latest/arednGettingStarted/node_status.html#node-and-broadcast-messages

PR #512, opened 2020-04-10
bugfix: migrate the TP-LINK CPE610 v1 device to the ath79 branch

apcameron:
The device was not stable on the ar71xx branch.
This PR migrates it to the ath79 branch.


ae6xe (approved):
code inspection looks good.  Tested by user in forum.

PR #513, opened 2020-04-14
aredn: correct AR750 ports in readme

dman776:


ae6xe (approved):
This now matches the behavior of the device configuration.

PR #519, opened 2020-04-22
bugfix: encode special characters in tunnel comment field

### LINKED ISSUE #515: Scrub Tunnel contact data field
encode or prevent < > | etc...  characters in the tunnel contact field.

r1de:
catch the tunnel comment field in perlfunc's read_post_data function where certain characters are changed to the html character entity in order to not interfere with the config files.
Tested with all kinds of crazy characters such as: zdsfkh? <!@#$%^&*()> "quote"

Fixes #515


ae6xe (approved):
looks good

PR #524, opened 2020-04-30
bugfix: aredn firewall blocking traffic when using tunnel feature

### LINKED ISSUE #522: meshchat does not sync with other nodes: Mikrotik hAP ac lite over tunnel (tun) - 3.20.3.0
I have been running down an issue that some others and I locally have been noticing with meshchat, and pings, when the other nodes are over tunnels.  This latest firmware is the first time I have ever tried tunnels, which, in general, is working for me!

However, meshchat was not syncing, and I narrowed it down to 'curl' on the mikrotik node that is either running as tunnel host/client (or both), not being able to connect to the remote node.  Curl just times out after 5 seconds, as meshchat has it do.

I don't know enough about iptables, but I think by 'accident' I found a workaround.  The file /etc/local/mesh-firewall/01-tunnels has a line:

 iptables -A zone_vpn_input -j zone_vpn_REJECT

that if I comment this line out and reboot, pings and meshchat magically start to work?!?

Can anyone help me understand what this file does, is this a bug?

I

ae6xe:
fixes #522
tested by: Matthew KB9OIV <Matthew.annen@gmail.com>
tested by: Chris K3ADA <sutehk.cs@gmail.com>

Resolves 2 issues with tunnel iptable rules.  A rule needed to be
shifted down by 1 position in chain given upgrade to openwrt 19.07.
Reload of rules was not correctly retaining chain order and creating
duplicate entries, inadvertantly blocking intended traffic.

### DIFF
--- a/files/etc/local/mesh-firewall/01-tunnels
+++ b/files/etc/local/mesh-firewall/01-tunnels
@@ -1,6 +1,7 @@
 #!/bin/sh
 <<'LICENSE'
   Part of AREDN -- Used for creating Amateur Radio Emergency Data Networks
+  Copyright (C) 2020 Joe Ayers
   Copyright (C) 2015 Conrad Lara and Joe Ayers
    See Contributors file for additional contributors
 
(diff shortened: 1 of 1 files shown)


r1de:
nice find!

PR #525, opened 2020-05-01
ath79: Migrate the TP-LINK WBS Devices to the ath79 Branch

apcameron:
Migrate the WBS devices to the ath79 branch

Add Support for WBS210-v2 and WBS510-v1

NOTE: I have only tested on my WBS510-v2 as I do not have the other devices.


dman776 (approved):
looks good

PR #529, opened 2020-05-26
openwrt: bump to 19.07.03

apcameron:
Remove 002-ar71xx-mikrotik-improved-support.patch
	Update 702-enable-country-hx.patch


ae6xe (approved):
Looks good.

k1ky:
Is this an Openwrt upgrade for all models or just the Microtik listed in the notes? In other words, is it worth updating/testing all of our different node/models from Nightly 1546 to 1566?

dman776:
Since it is an upgrade to openwrt, the more testing we can get on, the better

Thanks,
Darryl

ae6xe:
Just in case, yes, this is for all models, not solely for microtik.   We were able to remove an AREDN patch that is now available in the openwrt release we upgraded to.  One of our patches needed to go in a different order.  Other than that, no code changes were necessary to upgrade.   This openwrt upgrade and other AREDN fixes/updates since 3.20.3.0 suggest we should consider and are close to a 3.20.3.1 release.   Community testing will help us confirm stability to proceed.

PR #531, opened 2020-05-29
aredn: add Mode ether to tunnel interfaces

ae6xe:
This reduces message forwarding by OLSR. Without this mode
olsr will forward a message backout the same interface it
was received on, presumably due to hidden 802.11n nodes.

### DIFF
--- a/files/usr/local/bin/olsrd-config
+++ b/files/usr/local/bin/olsrd-config
@@ -144,6 +144,7 @@ if ( @tunnels )
   foreach(@tunnels) { push @file, qq("$_" ) }
   push @file, qq(\n{\n);
   push @file, qq(     Ip4Broadcast 255.255.255.255\n);
+  push @file, qq(     Mode \"ether\"\n);
   push @file, qq(}\n);
 }
 


ae6xe:
manually tested code change on hap ac lite.  Inspection of /var/run/olsrd.conf confirms setting is in use and "/etc/init.d/osrd restart" successfully starts up olsrd.

dman776:
in /www/cgi-bin/tunfunc.pm there is a add_olsr_interfaces().   shouldn't this be added there?
It's called when the user hits "Click to install" tunnels on the /vpn and /vpnc pages

ae6xe:
I couldn't find any location this routine was called -- appears to be unused and duplicate of the code in /usr/local/bin/olsrd-config.

ae6xe:
olsrd config is updated after "save" on the tunnel pages to add/remove a new tunnel -- only has in olsrd.conf the tunnels that are currently live.  So I don't think it is needed at tunnel install time.

dman776:
Ok.  Approved

dman776 (approved):

PR #532, opened 2020-05-29
aredn: supporttool: change format of firewall reporting

ae6xe:
use iptable -L -v options to see chain order and enable
improved ability to troubleshoot firewall issues


dman776 (approved):
Looks good

PR #541, opened 2020-06-08
ath79: CPE220V3 Fix

apcameron:
Remove Duplicate entry and fix remaining entry in the firmware file


ae6xe (approved):
Looks ok, I don't have a device to test on.

berniemcintosh:
If you confirm when it is in the latest nightly I will test. Thanks for your work on this  Bernie

shown19:
Hi about the CPE220v3 OpenWrt, there's a bug when all the connection drops -- "Obtaining IP Address", so I need to unplug and plug back the rj45 and the connection is back again. Yep, I don't need to reboot and the connection just fixed after replugging back the rj45. Tried with other cables so it's not the cable problem. I hope this can be resolved. 

Here's my post from the forum also: https://forum.openwrt.org/t/tl-cpe220v3-unstable-link/58648

PR #543, opened 2020-06-21
 aredn: remove node mode configuration notation

### LINKED ISSUE #533: Remove old configuration display
Remove reference on main UI page to "configuration mesh" - it's always mesh and no longer configurable anyway

### LINKED ISSUE #535: Revise Help file - remove reference to Client configuration
If AREDN github entry #533 is accepted

On Help Page 
Under Status Page -

**Change**

 The Signal/Noise/Ratio is a reading of the strongest neighbor Mesh RF signal strength (in dBm) from all connected stations, and it is available only when the node is in a Mesh or Client configuration. 
 
**To**
The Signal/Noise/Ratio is a reading of the strongest neighbor Mesh RF signal strength (in dBm) from all connected stations.

ae6xe:
In the past nodes could be configured uniquely for
a given mode:  ap, wifi client, mesh. Current behavior
in basic setup no longer has unique modes, rather wireless
cards can be configured in different modes in basic setup
and devices with multiple wireless interfaces can
be setup to use multiple modes at the same time.
   
fixes #533
fixes #535

### DIFF
--- a/files/www/cgi-bin/status
+++ b/files/www/cgi-bin/status
@@ -260,7 +260,7 @@ if($config eq "mesh" and ! $wifi_disable )
     push @col2, $str;
 }
 
-push @col2, "<th align=right><nobr>firmware version</nobr><br>configuration</th><td><nobr>" . `cat /etc/mesh-release` . "</nobr><br>$config</td>";
+push @col2, "<th align=right><nobr>firmware version</nobr></th><td>" . `cat /etc/mesh-release`. "</td>";
 push @col2, "<th align=right>system time</th><td>" . `date +'%a %b %e %Y<br>%T %Z'` . "</td>";
 
 $uptime = `uptime`;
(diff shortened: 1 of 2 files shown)


files/www/cgi-bin/status
```diff
@@ -260,7 +260,7 @@ if($config eq "mesh" and ! $wifi_disable )
     push @col2, $str;
 }
 
-push @col2, "<th align=right><nobr>firmware version</nobr><br>configuration</th><td><nobr>" . `cat /etc/mesh-release` . "</nobr><br>$config</td>";
```
dman776, line 263: Do we set $config on this page?   If so, can we remove it as well?

dman776 (approved):
Looks ok

files/www/cgi-bin/status
ae6xe: It is still used to detect when the node is not yet configured as pulled from an 'nvram' call, and displays the setup instructions accordingly.  While this value could be removed, and figure out what else to key off of,  it would be a super-set change affecting several files.  suggest we do separately or pull out with UI re-write.

PR #545, opened 2020-06-28
aredn: remove config files no longer in use consuming flash

ae6xe:
clean up 176k of flash space.  Config files include modes
forA P, Client, Mesh_Ap, and ROuter.  Files are no longer
referenced or used.


dman776 (approved):
Great catch.

PR #551, opened 2020-08-18
aredn: extend 5GHz channel options from 133 down to 131

### LINKED ISSUE #550: 5655 MHz channel 131 at 10 MHz bandwidth?
5, 10, and 20 bandwidths are selectable at 5920 MHz (channel 184).
Please add the selection of channels down to 130 MHz at 5 MHz bandwidth,
channel 131 at 10 MHz bandwidth, and channel 132 at 20 MHz bandwidth.
In our area of West Central Ohio, we know the location of the TDWR station.
The published frequency of that station is 5640 MHz (channel 128).
We would be using PtP links with directional antennas (dishes) and would
avoid radiating energy in the direction of the TDWR station.
I feel that we could operate at the above frequencies/bandwidths without causing interference.

Regards, Chuck nc8q

ae6xe:
In the US, part 97 ham radio lower band edge is 5650.
Channel 131 @ 10MHz channel width is the lowest channel
setting that stays within allocation.

fixes #550


ae6xe:
testing results in forum post:  https://www.arednmesh.org/comment/15858

dman776 (approved):
looks good

PR #553, opened 2020-08-18
aredn: add new Mikrotik model string SXTsq5HPnD

ae6xe:
Mikrotik change of string used to identify device

### DIFF
--- a/patches/009-add-sxtsq.patch
+++ b/patches/009-add-sxtsq.patch
@@ -2,7 +2,7 @@ Index: openwrt/target/linux/ar71xx/base-files/lib/ar71xx.sh
 ===================================================================
 --- openwrt.orig/target/linux/ar71xx/base-files/lib/ar71xx.sh
 +++ openwrt/target/linux/ar71xx/base-files/lib/ar71xx.sh
-@@ -575,6 +575,18 @@ mikrotik_board_detect() {
+@@ -577,6 +577,19 @@ mikrotik_board_detect() {
  	*"SXT Lite5")
  		name="rb-sxt5n"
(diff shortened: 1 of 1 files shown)


ae6xe:
ref this forum post of support data for new device and relevant dialog.
https://www.arednmesh.org/content/new-sxtsq-5-flash-issues

patches/009-add-sxtsq.patch
```diff
@@ -15,7 +15,8 @@ Index: openwrt/target/linux/ar71xx/base-files/lib/ar71xx.sh
 +	*"RBSXTsq2nD")
 +		name="rb-lhg-5nd"
```
dman776, line 16: should this be:  name="rb-lhg-2nd" ??

dman776 (commented):
see comment in code

patches/009-add-sxtsq.patch
ae6xe: This section is untouched before/after in this commit.    All of the SXT 2 and 5 devices function with the same hardware specification, so when they were first added, I did not create new board ids and reused the lhg-5nd.  The settings are otherwise detected of the model string.  At some point,  probably going into ath79 target, we should be able to inherit all of the definitions from upstream openwrt, and this will go away.  We had added these models before openwrt.

ae6xe: note the pre-existing sxt board ids were earlier hardware revisions that are not currently supported (the octagon form factor instead of the newer square form factor devices).

dman776 (approved):
Ok, makes sense

PR #555, opened 2020-08-21
bugfix: correct lower-case translation for hostname

dman776:
correct issue of case for hostname lookup for local alert files


ae6xe (approved):
Looks good.

PR #556, opened 2020-08-24
Change AdvConfig Warning Banner

ab7pa:
Change the Advanced Configuration warning banner: slightly wider to 
distinguish from Alerts banner, add WARNING and make text color pure 
white for contrast.


ae6xe (approved):
tested on an hap ac lite, all looked good.

PR #558, opened 2020-08-25
Add CSS fontfamily list

ab7pa:
Add fontfamily list in CSS files to help browsers render sans-serif 
correctly. Add text color so alert messages display correctly on all 
themes.


ae6xe (approved):

ae6xe:
Tested commit on hap ac lite.   Navigation though mesh status and node status with all theme option combinations worked as expected.

PR #560, opened 2020-08-26
enhancement: change advancedconfig boolean control

### LINKED ISSUE #429: Mikrotik hAP AC Lite POE passthrough advanced setting is confusing
The advanced setting for POE Passthrough (aredn.@poe[0].passthrough) is confusing because it does not conform to standard checkbox usage which states: 

> With a check box, users make a decision between two clearly opposite choices. The check box label indicates the selected state, whereas the meaning of the cleared state must be the unambiguous opposite of the selected state.

The initial setting is a checkbox that was not checked and the label says "OFF". That implies that you check the box to turn off POE passthrough, and that with it unchecked, POE is ON. After checking the box it then showed a check mark and the label says "ON". The label (or what appears to be the label) should not change to an opposite state depending on whether or not the checkbox is checked. After toggling the check It now is obvious that the label is being used to show that state of the checkbox, but that is what the check mark is for, not the label for the check. You could either remove the label, in which case the Config Setting name would substitute for it, or you could just leave the label as "ON" and the checked state would indicate whether or not it is on. This issue applies to USB passthrough, too, and maybe other settings, too.

dman776:
change widget rendering of boolean values on advancedconfig page to be more clear
resolves #429

### DIFF
--- a/files/www/cgi-bin/advancedconfig
+++ b/files/www/cgi-bin/advancedconfig
@@ -138,12 +138,14 @@ push @setting, {
   key => "aredn.olsr.restart",
   type => "boolean",
   desc => "Value doesn't matter.  Will restart OLSR when saving setting -- wait up to 2 or 3 minutes to receive response.",
+  default => "0",
   postcallback => "olsr_restart()"
 };
 push @setting, {
   key => "aredn.aam.refresh",
   type => "boolean",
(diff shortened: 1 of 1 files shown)


ae6xe (approved):
tested on hap ac lite, worked as advertised.

PR #561, opened 2020-08-27
Stop certain hostname/IP's from propagating over the mesh

### LINKED ISSUE #508: Ability to "hide" local hosts on Mesh Status page
I am suggesting a way to hide local services so that they do not appear on the Mesh Status page.  Here are a couple screen captures from my hAP at home that will explain this better.  First the DHCP reservations.  I have reservations for five WiFi gateways on my home WiFi network.  Currently that VLAN is not connected to any of the APs  Therefore, those IPs are not even in use - but there are reservations.  And therefore, they appear on the Mesh Status page.  Even if they were connected, there is zero reason for them to show up.
http://k6ccc.org/aredn/AREDN_DHCP-Res_page.png
http://k6ccc.org/aredn/AREDN_Local-hosts_page.png
My suggestion would be to add a checkbox for each reservation that hides that item from the Mesh Status page.  Additionally, add a checkbox that hides any attached device that has a DHCP
lease that does not have a reservation.  For example, if I plug my laptop into a node, there is generally no need for that laptop to show up on the Mesh Status page.
This would give the advantage to shortening the Mesh Status page to only useful information.
This is some related to issue #376

r1de:
Allows for the node admin to choose to have DHCP leased hostnames/IP's propagated over the rest of the mesh network or not.
Defaults to allowing the hostname/IP to propagate.

The hostname/IP will *still* be resolvable from the localnode and will show up in the list of hosts on the localnode only.
This allows for selected local mesh devices to be not available over the rest of the mesh network.
ie: switches, routers, cameras, etc.

This will work immediately for *new* DHCP leases when the checkbox is selected.
For *existing* DHCP leases, it may take a while for the network to update, if ever.
To speed up the process of full network OLSR "DNS" updating, reboot all the nearest neighbor device(s) to the node you made these changes to.
That seems to get the changes "out" to the rest of the network faster than normal.

Fixes #508

### DIFF
--- a/files/usr/local/bin/node-setup
+++ b/files/usr/local/bin/node-setup
@@ -290,14 +290,14 @@ foreach(`cat $dhcpfile`)
 {
   next if /^\s*#/;
   next if /^\s*$/;
-  ($mac, $ip, $host) = split /\s+/, $_;
+  ($mac, $ip, $host, $noprop) = split /\s+/, $_;
   $ip = decimal2ip($netaddr + $ip);
 
   # filter out addresses that are illegal for the lan subnet
   next unless validate_same_subnet($ip, $cfg{lan_ip}, $cfg{lan_mask});
   next unless validate_ip_netmask($ip, $cfg{lan_mask});
 
-  printf ETHER "$mac\t$ip\n";
+  printf ETHER "$mac\t$ip $noprop\n";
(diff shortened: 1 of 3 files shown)


ae6xe (approved):
tested on hap ac lite.  Looks good.

PR #562, opened 2020-08-29
Tweak aredn_message display

ab7pa:
Tweak the node identifier on the AREDN message display.


ab7pa:
I've been playing with Local Alerts for a while now and something like this seemed a little easier to read.  Local vs AREDN alerts are distinguished by the top header for each.
localmsg-display
The arrow symbol indicates "message to" followed by the destination node.

ae6xe (approved):
good changes, more descriptive of who is receiving the message.

PR #563, opened 2020-08-31
Add file size tests to arednMsg

ab7pa:
Add tests for file content to aredn_message.sh
There are cases for my local message repository (and probably for others) where it is normal to expect that an empty message file might exist.  I found that if an empty message file exists, then an empty message prefix/header is displayed.  I added a couple of file size tests to skip processing if the file has no meaningful content.  It may never occur, but I figured it wouldn't hurt to add tests to the AREDN Alerts as well as the Local Alerts.  Tested and verified that empty files in the local message repository will not be processed.


ae6xe (approved):
Looks good, improved logic adding && after wget command, to fall out if this command fails, and detection of empty download file if wget succeeds.

PR #565, opened 2020-09-07
aredn: bugfix max tunnel limited to 9 instead of 10

### LINKED ISSUE #564: Tenth tunnel will not activate for tunnel server (and probably 10th for client as well).  Bug in setup init script?
I got to configuring all 10 of the tunnel server slots today, and discovered that my final tunnel entry, the tenth, would not activate and become live when tested.

After confirming that both ends of the client/server were configured to match up, I 'swapped' the 9th and 10th tunnels.  It was in this case that the tunnel client that WAS in the 9th position on the server and working, was now not working in the 10th position, and the client that WAS in the 10th position (but now the 9th) WAS working.  Network numbers were changed appropriately on both ends in both cases, to match.

I took a closer look at the init script that sets up the tunnels on the server:

/etc/init.d/vtundsrv

and at line 53, I think there is an error in the comparison that is done for the loop that sets up each tunnel.

That line currently uses a 'less than' for the TUNNUM and MAXTUNNUM comparison:

TUNNUM=50
MAXTUNNUM=59

like this:

if [ $TUNNUM -lt $MAXTUNNUM ]

So I believe that when the final loop is done for the 10th tunnel, vtun59, the configuration is not completed, since in this case, TUNNUM is NOT less than MAXTUNNUM.-- 59 is not less than 59.

I changed the comparison to use 'less than or equal':

if [ $TUNNUM -le $MAXTUNNUM ]

and now the 10th tunnel activates.

This must be a small bug?

Finally, I do not have 10 tunnels on this device as client tunnels to test, but I suspect that the client setup will also show the same issue:

/etc/init.d/vtund

at line 71 the same comparison is done, with initial TUNNUM and MAXTUNNUM as 60 and 69, similarly.

ae6xe:
fixes #564

### DIFF
--- a/files/etc/init.d/vtundsrv
+++ b/files/etc/init.d/vtundsrv
@@ -50,7 +50,7 @@ allowed_client_config() {
 	local new_file="$2"
 	local enabled node pwd netip clientip serverip persist compress keepalive proto
 
-	if [ $TUNNUM -lt $MAXTUNNUM ]
+	if [ $TUNNUM -le $MAXTUNNUM ]
 	then
 		config_get_bool enabled "$cfg" enabled
 		config_get node "$cfg" node


dman776 (approved):
Good catch

PR #566, opened 2020-09-08
aredn: Mesh host aliases

### LINKED ISSUE #516: Enhancement Request:  Add DNS Aliases
Add the ability to make DNS aliases configured on a node be resolvable on nodes throughout the mesh.  With this feature, the same DHCP (or even static IP) host on an a node's LAN can be reached by names other than the host's hostname.  The DNS aliases should also be available as a dropdown when making services.

Example:  I have a Linux web server on my node's LAN.  I'd like to set up virtual servers for different purposes on that web server, each with it's own name.  The virtual server's name would have a DNS alias I'd configure on my node, and then those aliases would be resolvable on any other node on the mesh and be visible as a service on the other nodes.

Thank you,

Steve
AG7GN

r1de:
Allow for "CNAME" like aliases on the mesh.
The aliases can also be used to create propagated services using the aliased hostname.
Create new aliases on the DHCP and Ports page.
Once an alias is created it can be used to host a service, just choose the alias name in the service host dropdown.
The aliases will resolve _after_ the real hostname, so returns on pings may show a different host other than the alias, which is normal.
There is no limit for the number of aliases per IP, don't go nuts tho.
This can be used for virtual domain email servers, virtual hosts, virtual web sites, etc...

Fixes #516

### DIFF
--- a/files/etc/arednsysupgrade.conf
+++ b/files/etc/arednsysupgrade.conf
@@ -7,6 +7,7 @@
 /etc/config.mesh/_setup.ports.nat
 /etc/config.mesh/_setup.services.dmz
 /etc/config.mesh/_setup.services.nat
+/etc/config.mesh/aliases
 /etc/config.mesh/vtun
 /etc/config.mesh/aredn
 /etc/dropbear/dropbear_dss_host_key
--- a/files/usr/local/bin/node-setup
+++ b/files/usr/local/bin/node-setup
@@ -192,6 +192,7 @@ $portfile  = "/etc/config.mesh/_setup.ports";
 $dhcpfile  = "/etc/config.mesh/_setup.dhcp";
 $portfile .= ($cfg{dmz_mode} ? ".dmz" : ".nat");
 $dhcpfile .= ($cfg{dmz_mode} ? ".dmz" : ".nat");
+$aliasfile = "/etc/config.mesh/aliases";
 
 
 # basic configuration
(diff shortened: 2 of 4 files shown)


r1de:
Three other users and I have already tested this and it is working fine. We are all happy.

ae6xe:
@Eric KG6WXC did anyone test the alias PR after turning on LAN NAT mode?  Just thinking through, but the aliases should probably not be propagated or allowed.  The IP addresses aren’t reachable behind the NAT .  I didn’t get time to install and test tonight.

@Joe AE6XE  thought of that, and I tried it in NAT mode and realized that that it was pointless, so, as far as I could tell, the aliases just should not be there if you are in NAT mode.  I can check it again, but I tried to make it so that it didn't even show up in NAT mode.

@Eric KG6WXC  There might be a use case where someone with NAT mode would want an alias for their local use?   What about forcing a "no propagate" option if in NAT mode.    when someone checks the box to not propagate, NAT or no-NAT isn't involved with the usage at that point.    If forcing no propagate in NAT mode is minor effort to do, should try to get in this commit.  If much bigger effort, then can defer to another PR later.  thoughts?

r1de:
@ae6xe I get what you are saying, so user could create the alias for the internal NAT'd network and then use the alias in one of the port forwarding rules or something... I did not think of that scenario.  
I will have to look how NAT is working "under the hood", as I have never used that mode, I'd want to make sure it all goes together correctly, "no propagation" included.  
I think _I_ would want to wait and use a new PR for that, just so I have time to look at it fully and not feel pressed.

ae6xe:
multiple people tested in @r1de 's local group.

PR #569, opened 2020-09-10
aredn: visualize non-propagated and aliased hosts

r1de:
Show aliased and non propagated hosts differently on the mesh list.  
Only effects the localnode where the hosts are hidden/aliased.  
Just tries to provide a way for the node admin to easily visualize the setup from the `mesh` page.


r1de:
This is what it looks like in all the different styles, the colors are in CSS:  
CSS_Colors

dman776 (approved):

PR #573, opened 2020-09-25
aredn: enhancement: Nat mode aliases

r1de:
Added aliases in NAT Mode.  
NAT Mode aliases work much differently than normal DMZ Mode aliases, they are by default _NOT_ propagated out across the mesh, and cannot be made to do so, they _ONLY_ effect the node local LAN.  
You _CANNOT_ use an alias name in a service listing or a port forwarding rule, it's just not possible due to how `iptables` works.  
You _will_ be able to set an alternate hostname for any host on the nodes' LAN however.  
A host named `CBY45-DELLLAPTOP` can also be known as `wxc-shack-laptop`. It may make remembering which host is which a bit easier.  
  
Also snuck in a change to hide the "Do Not Propagate" checkbox while in NAT Mode.

### DIFF
--- a/files/etc/arednsysupgrade.conf
+++ b/files/etc/arednsysupgrade.conf
@@ -7,7 +7,8 @@
 /etc/config.mesh/_setup.ports.nat
 /etc/config.mesh/_setup.services.dmz
 /etc/config.mesh/_setup.services.nat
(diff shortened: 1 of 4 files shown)


ae6xe (approved):
Looks good

PR #576, opened 2020-09-26
bugfix: custom download paths during firmware upgrade

### LINKED ISSUE #575: bug: custom download paths
If custom download paths for packages are set, the custom values will _not_ be properly carried over when the devices firmware is upgraded or reflashed, even if "Save Settings" is checked.  
The custom values _are_ brought over via UCI configs, but the `/etc/opkg/distfeeds.conf` file still holds the default values after a firmware load.  
This then requires the device admin to go back to the adv. config page and save each custom setting again if they want the custom locations to work.  
`opkg` only reads the `/etc/opkg/distfeeds.conf` file when looking for the location to download packages from, it does not look in the UCI configs.  
If there are custom URL's they should be carried over during upgrades as with other settings.  
  
(This only applies to the package URL's, not the firmware download URL.)

r1de:
Fixes #575

### DIFF
--- a/files/etc/init.d/local
+++ b/files/etc/init.d/local
@@ -89,9 +89,8 @@ boot() {
       uci -c /etc/config.mesh commit aredn
     # check values in distfeeds.conf against UCI settings
     # and change distfeeds.conf if needed (upgrades?)
-    ## commented for now, may need it later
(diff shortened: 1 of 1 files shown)


ae6xe (approved):
Looks good

PR #580, opened 2020-10-07
aredn: add new mikrotik model string for LHG based devices

ae6xe:
new model string: MikroTik RouterBOARD RBLHG-5HPnD


ae6xe:
tested by multiple users.  See this forum thread:  https://www.arednmesh.org/comment/16146#comment-16146

dman776 (approved):
tested previously ok

PR #587, opened 2020-12-07
ath79: Enable the LNA for ALL for the TP-Link CPExxx AR9344 Devices

apcameron:
This should improve the Power Levels for the 1 Port Devices as it was not enabled for them.


ae6xe (approved):
looks good.

PR #588, opened 2020-12-10
openwrt: bump version to v19.07.4

apcameron:
Bump the Openwrt version to v19.07.04 as the sources for the previous version are missing in some cases.


ae6xe (commented):
19.07.5 is officially released in the last ~day or so.  There are some CVEs in the kernel and at least one applicable bug since 19.07.4:
Description: GL-AR300M: Frequent WAN (eth1) link down / link up
Link: https://bugs.openwrt.org/index.php?do=details&task_id=3226

Shouldn't we go to 19.07.5 instead?

dman776:
let me test 19.07.5 now with Andrew's patch...

apcameron:
From what I can see the patches apply to 19.07.5. If it compiles for @dman776 I will update the PR to use 19.07.5

dman776:
It built ok

PR #594, opened 2021-03-26
Update perlfunc.pm

kc8ufv:
Added entries for all Mikrotik Routerboard models with RB added before the number to support their changing that in the hardware. (I ignored the ones that were already done, such as the 952)


ae6xe:
PRs should be submitted to the https://github.com/aredn/aredn repo.  This repo is frozen.   Also, please see this commit.  Need to check if one more change is needed -- 2 places to update.  The image may not load.

https://github.com/aredn/aredn/commit/bf944dface44f1271998a77b33cbd89c0393d099

kc8ufv:
Sorry, put it to the wrong one. Realized it after submitting. Already
resubmitted to aredn/aredn.

